PR #49, opened 2021-06-09
Namespace authorization updates

### LINKED ISSUE #46: Namespace name validation

### LINKED ISSUE #47: Namespace Selector should check NS with Namespace.View or Namespace.Manage
Currently, the Namespace selector only shows Namespaces if the user has the `Namespace.Manage` permission.  It should be selectable if the user has `Namespace.View` as well.

### LINKED ISSUE #50: User with no namespaces showing error

ikethecoder:
Resolves #46 
Resolves #47 
Resolves #50


src/nextapp/components/new-namespace/new-namespace.tsx
```diff
@@ -38,9 +38,9 @@ const NewNamespace: React.FC<NewNamespace> = ({ isOpen, onClose }) => {
       if (form.current.checkValidity()) {
         try {
           const name = data.get('name') as string;
-          const json = await createMutation.mutateAsync({
+          const json = (await createMutation.mutateAsync({
             name,
-          }) as { createNamespace: { name: string; id: string }};
+          })) as { createNamespace: { name: string; id: string } };
```
Jonesy, line 43: I discovered there's actually an easier way to type these that saves you the inline assertion. Import the app types `import type { Mutation } from '@/types/query.types';`

and then 

`const res: Mutation = await createMutation.mutateAsync({ ... })`

PR #121, opened 2021-07-13
Improved handling of Public API and credential collection

### LINKED ISSUE #78: Update alert to reflect pending approval
Content on the API Requested Access page to be updated to ensure it describes the difference between "you now have access" vs "credentials will be authorized once approved".

ikethecoder:
Fixes #78 

Enhancements:
* Disables the "Request Access" button when there are no protected environments for the API
* On "My Access", when you mouse over the Pending/Approved Icon, show the Application that the access relates to
* Display requests that have credentials which have not been collected (and have option to cancel request)
* Don't show public flows in the authorization panel on Consumer detail page

### DIFF
--- a/src/nextapp/components/access-requests/access-requests.tsx
+++ b/src/nextapp/components/access-requests/access-requests.tsx
@@ -104,7 +104,7 @@ const AccessRequests: React.FC = () => {
             </Tr>
           )}
           {data?.allAccessRequestsByNamespace
-            .filter((req) => req.serviceAccess != null)
+            .filter((req) => Boolean(req.serviceAccess))
             .slice(0, sliceIndex)
             .map((d) => (
               <Tr key={d.id}>
--- a/src/nextapp/components/access-requests/access-requests.tsx
+++ b/src/nextapp/components/access-requests/access-requests.tsx
@@ -32,6 +32,9 @@ const query = gql`
         name
         username
       }
+      serviceAccess {
+        id
+      }
     }
   }
 `;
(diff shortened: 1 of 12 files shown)


src/nextapp/pages/devportal/requests/new/[id].tsx
```diff
@@ -142,7 +142,7 @@ const NewRequestsPage: React.FC<
       client.invalidateQueries('allAccessRequests');
       toast({
         title: 'Request submitted',
-        description: 'Check back to see if it has been accepted soon',
+        description: '',
```
Comment (Jonesy) [inline src/nextapp/pages/devportal/requests/new/[id].tsx:145]: You can just leave this prop off if there's no content.

src/nextapp/components/access-requests/access-requests.tsx
```diff
@@ -46,17 +49,14 @@ const AccessRequests: React.FC = () => {
     },
     { suspense: false }
   );
-  const total = data?.allAccessRequestsByNamespace.length ?? 0;
+  const total =
+    data?.allAccessRequestsByNamespace.filter(
+      (req) => req.serviceAccess != null
```
Jonesy, line 54: For comparisons like this it's actually more reliable to just go `.filter((req) => Boolean(req.serviceAccess))`

src/nextapp/pages/devportal/requests/new/tokens.tsx
```diff
@@ -53,20 +55,33 @@ const ApiAccessPage: React.FC = () => {
     router?.push('/devportal/access');
   };
 
-  return (
+  return access.data ? (
```
Jonesy, line 58: I don't really like this pattern for Next pages because it defeats the purpose of using SSR, and a lot of useful info like the head tags has to be re-rendered. I could demo a different approach, if you want to wait on additional content we can make some suspense components nested inside the first page render.

ikethecoder: Ok, I removed it and only included the logic in the part that I wanted to display only when there is real content to show.

PR #163, opened 2021-08-04
Fix to add Rate Limit Plugin with Redis Policy

### LINKED ISSUE #104: Consumer Rate Limiting by Redis not working
Setting up rate limiting by Redis requires the plugin to be setup with Redis credentials.  The Gateway Administration (gwa) API has these Redis credentials and therefore should be used to update the Consumer plugins rather than the Portal.  Perhaps any Consumer-related gateway changes should be moved to the gwa-api for consistency.

NithinKuruba:
Fixes #104


src/services/kong/consumer-service.ts
```diff
@@ -39,6 +39,7 @@ const logger = Logger('kong');
 
 export class KongConsumerService {
   private kongUrl: string;
+  private gwaUrl: string = process.env.GWA_API_URL;
```
ikethecoder, line 42: Can you make consistent the setting of the values for`kongUrl` and `gwaUrl`?  They both are set from environment variables.  I prefer constructor injection, but either way, approach should be same.

ikethecoder (changes requested):

src/services/kong/consumer-service.ts
NithinKuruba: @ikethecoder Yup, I like the idea of constructor injection. Made the changes.

ikethecoder (approved):

PR #182, opened 2021-08-22
Create Application From Access Request Page

### LINKED ISSUE #168: Create an application from Access Request page
# API Services Portal Issue

## User Story

**As a** Developer,<br>
**I want** to be able to create an application from the Request Access page<br>
**so that** I don't have to go back and forth between the Applications page and the API Directory page.

### ENV

- [X] `DEV`
- [X] `TEST`
- [X] `PROD`

### TESTCASE

- Select an API from the API Directory
- Click the "Request Access" button
- Have a button that can be clicked to "Add new Application" that will add an application and backfill it into the Application dropdown

NithinKuruba:
Fixes #168

### DIFF
--- a/src/nextapp/components/new-application/new-application-dialog.tsx
+++ b/src/nextapp/components/new-application/new-application-dialog.tsx
@@ -23,6 +23,7 @@ interface NewApplicationDialog {
   open: boolean;
   onClose: () => void;
   userId: string;
+  refreshQueryKey: string;
   handleAfterCreate?: (app: Application) => void;
 }
 
@@ -30,6 +31,7 @@ const NewApplicationDialog: React.FC<NewApplicationDialog> = ({
   open,
   onClose,
   userId,
+  refreshQueryKey,
   handleAfterCreate,
 }) => {
   const toast = useToast();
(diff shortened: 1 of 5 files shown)


src/nextapp/components/new-application/new-application-dialog.tsx
```diff
@@ -49,7 +54,7 @@ const NewApplicationDialog: React.FC<NewApplicationDialog> = ({
         if (form.current.checkValidity()) {
           const name = data.get('name') as string;
           const description = data.get('description') as string;
-          await applicationMutation.mutateAsync({
+          const a: Mutation = await applicationMutation.mutateAsync({
```
Jonesy, line 57: Nit picky but I would use a more descriptive variable name here, I tend to only use single letter const for iteration handlers since the variable you're looping over is defined at the start.

NithinKuruba: @Jonesy hahah I missed to update that. I put that for testing and now updated it.

src/nextapp/components/new-application/new-application-dialog.tsx
```diff
@@ -49,18 +54,22 @@ const NewApplicationDialog: React.FC<NewApplicationDialog> = ({
...
+          );
           toast({
             title: `${name} created!`,
             description: 'You can now request access to an API',
             status: 'success',
           });
-          queryClient.invalidateQueries('allApplications');
+          queryClient.invalidateQueries(refreshQueryKey);
           onClose();
+          if (handleAfterCreate)
```
Jonesy, line 71: Last nitpicky thing I promise! Just wrap conditionals with `{ ... }`, with JavaScript it's best practices and code consistency. Otherwise it's looking good!

NithinKuruba: @Jonesy I added them

PR #247, opened 2021-11-30
Update 01-create-api.spec.ts for following changes:

nirajCITZ:
1)Remove steps to select the value from Organization and Organization Unit for Edit Product page
2)Update action to select Term and condition and Kong Config from the drop down in Edit Environment
3)Update step to verify user name in Sign Out button

### DIFF
--- a/e2e/cypress/pageObjects/logout.ts
+++ b/e2e/cypress/pageObjects/logout.ts
@@ -1,0 +1,7 @@
+class LogoutPage {
+  path: string = '/'
+
+  logoutButton: string = '[data-testid=auth-menu-user]'
+}
+
+export default LogoutPage
--- a/e2e/cypress/pageObjects/products.ts
+++ b/e2e/cypress/pageObjects/products.ts
@@ -26,8 +26,8 @@ class Products {
   editProduct(productName: string, orgName: string, orgUnitName: string) {
     const pname: string = productName.toLowerCase().replaceAll(' ', '-')
     cy.get(`[data-testid=${pname}-edit-btn]`).first().click()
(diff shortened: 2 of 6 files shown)


e2e/cypress/pageObjects/products.ts
```diff
@@ -40,11 +40,12 @@ class Products {
...
     cy.get(this.envCfgApprovalCheckbox).click()
-    cy.get(this.envCfgTermsDropdown).select(config.terms)
-    cy.get(this.envCfgAuthzDropdown).select(config.authorization)
+    cy.get(this.envCfgTermsDropdown).select(config.terms,{ force: true }).invoke('val')
+    cy.get(this.envCfgAuthzDropdown).select(config.authorization,{ force: true }).invoke('val')
     cy.get(this.envCfgOptText).type(config.optionalInstructions)
-    cy.get(`[data-testid=${config.serviceName}`).click() //Adding service to list of active services
+    //cy.get(`[data-testid=${config.serviceName}`).click() //Adding service to list of active services
     cy.get(this.envCfgApplyChangesBtn).click()
+    cy.wait(10000)
```
NithinKuruba, line 48: @nirajCITZ Please avoid using `wait` statements as they tend to cause unexpected behaviors.

e2e/cypress/pageObjects/logout.ts
```diff
@@ -0,0 +1,7 @@
+class LogoutPage {
```
NithinKuruba, line 1: We do not have a logout page in portal. So I believe its not a good idea to represent logout component as a page

PR #272, opened 2021-12-14
Niraj local  1

nirajCITZ:
1)Make changes to make the script executable without cookie issue (Zenhub ticket # 255)
2)Added verification step to check that the API is accessible with the API key once the access request is granted


e2e/cypress/support/auth-commands.ts
```diff
@@ -207,6 +207,31 @@ Cypress.Commands.add('publishApi', (fileName: string) => {
...
+  var cookies = document.cookie.split(";");
+  for (var i = 0; i < cookies.length; i++) {
+      var cookie = cookies[i];
+      var eqPos = cookie.indexOf("=");
+      var name = eqPos > -1 ? cookie.substr(0, eqPos) : cookie;
+      document.cookie = name + "=;expires=Thu, 01 Jan 1970 00:00:00 GMT";
+  }
+})
+
+Cypress.Commands.add('getAPIRequest', (serviceName : string) => {
```
NithinKuruba, line 223: I recommend using a different name as `getAPIRequest` would not convey the right info about what it actually does. Infact, you can create a command that can handle all the requests like GET, POST/PUT and DELETE, could be used in upcoming tests. Probably call this function `makeKongRequest`?

e2e/cypress/support/auth-commands.ts
```diff
@@ -207,6 +207,31 @@ Cypress.Commands.add('publishApi', (fileName: string) => {
   })
 })
 
+Cypress.Commands.add('deleteAllCookies', () => {
```
NithinKuruba, line 210: @nirajCITZ we had a custom command `removeCookies`. so if its not being used anywhere, please feel free to remove it as it would be good to have just one function to clear the cookies.

NithinKuruba (changes requested):
@nirajCITZ I have added couple of comments for changes.

e2e/cypress/support/auth-commands.ts
nirajCITZ: Done

e2e/cypress/support/auth-commands.ts
nirajCITZ: Done

e2e/cypress/support/auth-commands.ts
```diff
@@ -207,6 +206,32 @@ Cypress.Commands.add('publishApi', (fileName: string) => {
...
+  var cookies = document.cookie.split(";");
+  for (var i = 0; i < cookies.length; i++) {
+      var cookie = cookies[i];
+      var eqPos = cookie.indexOf("=");
+      var name = eqPos > -1 ? cookie.substr(0, eqPos) : cookie;
+      document.cookie = name + "=;expires=Thu, 01 Jan 1970 00:00:00 GMT";
+  }
+})
+
+Cypress.Commands.add('makeKongRequest', (serviceName : string, methodType : string) => {
```
NithinKuruba, line 222: @nirajCITZ thanks for updating this function. The function need to accept headers and payload as well to make it work for POST/PUT requests. The headers may be added and updated depending upon different scenarios.

nirajCITZ: As discussed, we will keep optimizing it as we come across with different scenarios

NithinKuruba (approved):

PR #285, opened 2022-01-12
Niraj local

nirajCITZ:
Added Scenarios for IP Restriction and Rate limiting


e2e/cypress/pageObjects/consumers.ts
```diff
@@ -1,3 +1,79 @@
 export default class ConsumersPage {
   path: string = '/manager/consumers'
+  rateLimitHourInput: string = 'input#hour'
```
NithinKuruba, line 3: @nirajCITZ I recommend using testids as they cannot be altered by UI changes. Please update the same for rest of the declarations.

e2e/cypress/support/util-commands.ts
```diff
@@ -74,10 +71,32 @@ Cypress.Commands.add('getState', (key: string) => {
   }
 })
 
+Cypress.Commands.add('getPayload', (name: string) => {
+  let payload = {}
+  const connections = {
+    rate_limiting: {
+      "name": "rate-limiting",
+      "config.hour": "20",
```
NithinKuruba, line 79: @nirajCITZ please move the hardcoded values to a fixture, which would help us reuse the function other scenarios

e2e/cypress/tests/manage-control/10-rate-limiting.ts
```diff
@@ -0,0 +1,249 @@
...
+        cy.visit(login.path)
+    })
+
+    it('set api rate limit to 1 request per min, Local Policy and Scope as Service', () => {
+        cy.get('@access-manager').then(({ user, namespace }: any) => {
+            cy.login(user.credentials.username, user.credentials.password).then(() => {
+                home.useNamespace(namespace);
+                cy.visit(consumers.path);
+                consumers.clickOnTheFirstConsumerID()
+                consumers.setRateLimiting('1')
```
NithinKuruba, line 29: Same here, please move the hardcoded values to a fixture

e2e/cypress/pageObjects/consumers.ts
nirajCITZ: done...'data-testid' are not available but update all the elements that have ids properties in consumer page

e2e/cypress/tests/manage-control/10-rate-limiting.ts
nirajCITZ: Here, we are setting 1 request per hour so it does not need to move to fixture..Here we will arrange the test...Please let me know if I understood wrongly

NithinKuruba (approved):

PR #298, opened 2022-01-20
Authorization check (ACL Switch) scenario

nirajCITZ:
Added Scenarios for authentication check (Zenhub#292)


e2e/cypress/pageObjects/consumers.ts
```diff
@@ -1,81 +1,72 @@
...
-  //   cy.contains('button','Edit').click()
-  // }
-
-  deleteControl()
-  {
-      cy.get('body', {log: false}).then(($body) => {
-        if($body.find(this.removeIPRestrictionButton).length > 0) {
-          cy.get(this.removeIPRestrictionButton).first().click()
-          cy.contains('button','Yes, Delete').click()
+  turnOnTheSwitch(flag: Boolean) {
```
jTendeck, line 53: Maybe a bit more of a descriptive name for this function. What switch is being turned on? It's not so apparent when this function is being called outside of this file

PR #321, opened 2022-02-08
Update DEV with changes from Automation Branch for this Sprint

NithinKuruba:


ikethecoder (commented):
Does there need to be a separate Dockerfile.E2E?  For example, a new script like "start-with-coverage" in package.json that calls `nyc instrument --compact=false nextapp --in-place` before calling `start`?  And then override the CMD to be "start-coverage" when running the docker container.  If it is needed, perhaps move it to the `e2e` folder.

NithinKuruba:
> @NithinKuruba I spied a couple debugger statements in the test files, are they inconsequential for the Cypress tests or should they be removed before merging?

Yep, I will remove them

PR #482, opened 2022-08-05
Quick fix for when data is not prefetched on documentation page

### LINKED ISSUE #480: Next direct links to documentation fails due to missing object
Links from https://www2.gov.bc.ca/gov/content/data/about-data-management/databc/databc-products-services/api-management to the APS Portal website throw errors when visiting directly. They can be worked around by visiting the Documentation tab then visiting the link after.

Problem URL: https://api.gov.bc.ca/docs/platform-api-owner-user-journey

The error seems to be thrown by the document object being undefined while trying to access the `title` property.

Jonesy:
* Fixes #480 

It seems like the prefetch doesn't always return a data record on render, so this fix adds some guards incase this happens.

### DIFF
--- a/src/nextapp/pages/docs/[slug].tsx
+++ b/src/nextapp/pages/docs/[slug].tsx
@@ -80,22 +80,22 @@ const DocsContentPage: React.FC<
   return (
     <>
       <Head>
-        <title>API Program Services | Documentation | {data.title}</title>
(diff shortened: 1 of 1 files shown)


ikethecoder:
Not implementing as we have removed the documentation tab off of the Portal

Issue #494, opened 2022-08-10
Approver name incorrect on Consumer detail dialog


ikethecoder:
"Approver" is incorrect - should be the Request Approver, not the requester.

![image.png](https://images.zenhubusercontent.com/5c3e5c85cff96b450939aec6/384ced9c-fcc3-4daf-960c-2ae96484920b)

ikethecoder:
**Technical comments:** This is due to the way the Approver is derived from the Activity details.  Need to update the content of the Activity so that it is easy to determine which event is related to approval vs requester.

Issue #493, opened 2022-08-10
Business profile not appearing on the Access approval dialog


ikethecoder:
# API Services Portal Issue

## User Story

**As a** Access Manager,<br>
**I want** to see the business profile when the user is logging in with a Business-based identity<br>
**so that** I can verify what company they are requesting access for.

### ENV

- [ ] `DEV`
- [X] `TEST`
- [ ] `PROD`

### TESTCASE

- Request access to a product that has approval required
- Log in as an Access Manager for that product
- Go to Consumers and select for review the newly created access request

### EXPECTED

- The Business Profile to be populated

### ACTUAL

- Get a message "Unable to load business profile"

### ERROR

Getting an "ServiceAccessNotFound" error - most likely the Business Profile GraphQL call is getting sent the wrong ID.

PR #497, opened 2022-08-11
Safer log handling and security fix

ikethecoder:
Beyond better log handling and the security fix, the following was also resolved:
* Role for the user's identity provider was being incorrectly set when switching namespace
* Namespace directory api returning 204
* #493 
* #494


src/services/keystone/product-environment.ts
```diff
@@ -147,10 +147,10 @@ export async function lookupEnvironmentAndIssuerUsingWhereClause(
                 }`,
     variables: { where },
   });
-  logger.debug(
-    '[lookupEnvironmentAndIssuerUsingWhereClause] result %j',
-    result
-  );
+  // logger.debug(
```
Jonesy, line 150: Should we just remove these? Or will they come in handy down the road?

jTendeck (approved):

PR #594, opened 2022-10-08
fixes for updating roles in clients

ikethecoder:


src/services/keycloak/client-roles.ts
```diff
@@ -0,0 +1,161 @@
...
+            userRoles.filter((urole: any) => urole.name === role.name).length ==
+            0
+        )
+        .map((role) => ({ id: role.id, name: role.name }));
+
+      const delRoles = userRoles
+        .filter((urole: any) => selectedRoles.includes(urole.name) == false)
+        .map((role: any) => ({ id: role.id, name: role.name }));
+
+      changeList.push(...addRoles.map((r: any) => `Role Add ${r.name}`));
```
Jonesy, line 132: I *think* you could just chain at `.forEach` at the end of `addRoles` and `delRoles` that push the string to `changeList` without having to destruct and map over the results, it is a bit more CPU intensive to approach it this way.

Jonesy (approved):

PR #600, opened 2022-10-13
Service Accounts UI Refresh

### LINKED ISSUE #597: Service Accounts UI Refresh

Jonesy:
Implements the new Service Accounts UI, fixes #597


src/nextapp/pages/manager/service-accounts/index.tsx
```diff
@@ -2,32 +2,40 @@ import * as React from 'react';
 import api, { useApi } from '@/shared/services/api';
```
jTendeck, line 2: Might just need to run Prettier in this file as the formatting appears to be a bit off in places

Jonesy: Good catch! I also saw there was an unused variable and removed that as well.

PR #617, opened 2022-10-25
Add footer pages

jTendeck:


src/nextapp/pages/content/about-us.tsx
```diff
@@ -0,0 +1,74 @@
+import FooterPage from '@/components/footer-page';
+import Head from 'next/head';
+
+const AboutPage: React.FC = () => {
+  let pageHeader = 'About Us';
```
Jonesy, line 5: Use `const` over `let` if you are not going to mutate this value. I generally just write string props directly on the component.

src/nextapp/components/footer-page/footer-page.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { Box, Container, Text } from '@chakra-ui/react';
+import PageHeader from '@/components/page-header';
+
+interface FooterPageProps {
+  pageHeader: string;
+  updateDate: string;
+  content: string;
+}
+
+const FooterPage: React.FC<FooterPageProps> = (props) => {
```
Jonesy, line 13: Prefer the convention of destructuring prop arguments instead of a single `prop` argument. ie `const Component = ({ pageHeader, updateDate, content }) =>`

Jonesy, line 13: Generally try to stay away from super specific names like `FooterPage` for a layout that likely could be used for pages not coming from the Footer. Something like `ContentPage` would work better.

Jonesy (changes requested):
Use Next.js' Markdown support to render Markdown files instead of parsing strings inside React components.

src/nextapp/pages/_app.tsx
```diff
@@ -137,18 +133,20 @@ const App: React.FC<AppProps> = ({ Component, pageProps }) => {
...
-                          borderRight={{
-                            base: 'none',
-                            md: '1px solid #4b5e7e',
-                          }}
-                          px="5px"
-                          py={{ base: '5px', md: 0 }}
-                        >
-                          {f.text}
+                        <Link href={f.href}>
+                          <Box
```
Jonesy, line 137: One adjustment here, generally I prefer to import the UI kit's `Link` component and bring in Next's link as `NextLink`. Then structure the links like this

```tsx
<NextLink passHref href={f.href}>
  <Link fontSize="xs" px="5px" py={...}>
    {f.text}
  </Link>
</NextLink>
```
The reason is because Next Links aren't anchor tags by default, so we loose some accessibility for free, as well as semantic CSS. The Chakra `Link` component also gives us normal link CSS behaviour for free, so we don't have to add other classes like `cursor`, `display`, etc to mimic the standard behaviour.

Jonesy (changes requested):

src/nextapp/pages/_app.tsx
jTendeck: Ahh, thanks for the feedback and info Josh... I've implemented your feedback :)

Jonesy (approved):

PR #750, opened 2023-02-13
Feature Email Editing and Null User

### LINKED ISSUE #452: Collect email for notifications
**Story**: As a Developer that logs in with an Identity Provider that does not provide Email, I want to provide an email so that I can be notified when my request for access has been approved.

* [x] Design
* [x] Implementation

Test Scenarios:
1. BCSC user login for the first time and email updated: Confirm the user is prompted with dialog to update email, and that the email was updated
2. BCSC user login for the second time : Confirm no prompt for email and previously updated email appears in Profile
3. IDIR user login for the first time : Confirm user's email is set from Identity Provider and not prompted to change and no Edit email on profile
4. IDIR user login for the second time : Confirm no prompt for email
5. BCSC user login for the first time from the Request Access flow and is prompted for email : Confirm the user comes back to the Request Access page with the Email update dialog; and after update, remains on the Request Access page

### LINKED ISSUE #717: null Users
As a user who logged in using Github, I don't want to be referred as `null` on the emails. 

As an api provider looking at the activity feed, any other information other than user = `null` could be more helpful. 

DoD: find alternatives to replace `null`

ikethecoder:
* Closes #717 
* Closes #452

Feature branch testing: https://api-services-portal-feature-email-editing.apps.silver.devops.gov.bc.ca/


ikethecoder:
@Jonesy is it by design that you have to put the email in?  If I try to focus out of or click the "X" on top right, it does not close.

image

Issue #810, opened 2023-05-12
CKAN synchronization with BCDC is not working as expected


chrsamp:
# API Services Portal Issue

The list of organizations and business units to add to a namespace is not synchronized with the [list of organizations and business units](https://catalogue.data.gov.bc.ca/organization) provided by the BCDC. Organizations and business units are also different between TEST and PROD, but that's a less pressing issue.

## Steps to Reproduce
- Log into the portal as an API provider
- Create a new namespace or choose a namespace that is not matched to an organization
- Navigate to "Namespaces" then click "Add Organization"
- Note the choices do not align with the choices provided by the BCDC

## Test Case/Expected Behaviour
- When choosing "Ministry of Water, Land and Resource Stewardship" the business unit "Natural Resource Information and Digital Services" should appear

## Current Behaviour
- The list of business units differs significantly (also for other ministries, e.g. Municipal Affairs, Indigenous Relations and Reconcilation, etc.)

PR #837, opened 2023-06-16
Fix BCDC Datasets out of sync

ikethecoder:
Feature branch: https://api-services-portal-feature-bcdc-sync-fix.apps.silver.devops.gov.bc.ca/signout

### DIFF
--- a/.github/workflows/ci-build-feeders.yaml
+++ b/.github/workflows/ci-build-feeders.yaml
@@ -138,8 +138,6 @@ jobs:
               value: 'http://prometheus-server'
             DESTINATION_URL:
               value: 'http://proto-asp-${{ steps.set-deploy-id.outputs.DEPLOY_ID }}-generic-api'
-            SCHEDULE:
-              value: 'true'
           " > values.yaml
 
           helm repo add bcgov http://bcgov.github.io/helm-charts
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-build-feeders.yaml
```diff
@@ -138,6 +138,8 @@ jobs:
               value: 'http://prometheus-server'
             DESTINATION_URL:
               value: 'http://proto-asp-${{ steps.set-deploy-id.outputs.DEPLOY_ID }}-generic-api'
+            SCHEDULE:
+              value: 'true'
```
ikethecoder, line 142: Since this will be common for all feature branches, once merged into dev, this should be set to "false" - we don't want all the feature branch instances to be calling BCDC and Kong.  Sync can be run manually if the particular feature needs it.

feeds/index.js
```diff
@@ -171,6 +171,8 @@ if (process.env.SCHEDULE == 'true') {
   runTimedJob('prometheus', 120, { numDays: 1 });
   runTimedJob('prometheus', 24 * 60 + 5, { numDays: 5 });
   runTimedJob('kong', 1 * 60, {});
+  runTimedJob('ckan', 24 * 60, {});
```
ikethecoder, line 174: Perhaps run `ckan` a little bit later than `ckan_org` as it does reference organizations.  So maybe time is `(24*60) + 30`?

Elson9:
Fixes #810

PR #879, opened 2023-08-17
fixes request access button not being displayed when the Product has …

### LINKED ISSUE #631: 'Request Access' button is not displayed when the Product has an inactive environment
Step 1: Add an environment to the product
Step 2: Make the added environment inactive (do not configure environment or service details)
Step 3: Make sure that the product has some active environment
Step 4: Sign as API consumer 
Step 5: Select the Product where an inactive environment is added from API Directory tab
Step 6: Verify 'Request Access' button

Actual Result : 'Request Access' button is not displayed when the Product has an inactive environment

Expected Result : 'Request Access' button should be displayed but user is not allowed to select inactive environment while raising access request

Elson9:
…an inactive environment

Fixes #631

### DIFF
--- a/src/nextapp/components/api-product-item/api-product-item.tsx
+++ b/src/nextapp/components/api-product-item/api-product-item.tsx
@@ -63,15 +63,15 @@ const ApiProductItem: React.FC<ApiProductItemProps> = ({
             )}
           </GridItem>
         </Grid>
-        {!isPublic && !isTiered && (
+        {!isTiered && (
           <>
             {isPublic && (
-              <Button
-                rightIcon={isPublic ? <Icon as={BiLinkExternal} /> : undefined}
-                data-testid="api-product-try-button"
-              >
-                Try this API
-              </Button>
+              <AccessRequestForm
+                disabled={false}
+                id={id}
+                name={data.name}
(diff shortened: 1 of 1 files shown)


src/nextapp/components/api-product-item/api-product-item.tsx
```diff
@@ -63,15 +63,15 @@ const ApiProductItem: React.FC<ApiProductItemProps> = ({
             )}
           </GridItem>
         </Grid>
-        {!isPublic && !isTiered && (
+        {!isTiered && (
           <>
             {isPublic && (
```
Jonesy, line 68: Because you're rendering the same form component and props I think you can just render this component alone in the `!isTiered` check, it's the same as showing it if `isPublic` was true or false.

Jonesy (changes requested):

src/nextapp/components/api-product-item/api-product-item.tsx
Elson9: Thanks Josh, I had left that in originally in case there's a future need for that logic, but I guess we can just add it in if/when necessary.

ikethecoder (approved):

PR #917, opened 2023-09-20
Fixed namespace creation regex

### LINKED ISSUE #822: Namespace can be created with unsupported characters
# API Services Portal Issue

Relates to #806. A user can create a namespace that ends with a hyphen. This should not be permitted as you cannot publish the gateway config (using either `gwa-cli` or `gwa-api`) containing a tag ending with a hyphen.

### ENV

- [ ] `DEV`
- [x] `TEST`
- [ ] `PROD` (likely exists in `PROD`, I only verified in `TEST`)

### TESTCASE

- Create a new namespace ending in a hyphen (i.e. `apstest-`)

### EXPECTED

- The user cannot create the namespace and a validation error is displayed to the user.

### ACTUAL

- The namespace is created successfully.

## Notes
This is the sample yaml config I used to test whether I was able to publish:

```yaml
services:
- name: apstest-example2
  host: httpbin.org
  tags: [ ns.apstest- ]
  port: 443
  protocol: https
  retries: 0
  routes:
  - name: apstest-example-route2
    tags: [ ns.apstest- ]
    hosts:
    - apstest-example2.api.gov.bc.ca
    preserve_host: false
    paths:
    - /
    methods:
    - GET
    strip_path: false
    https_redirect_status_code: 426
    path_handling: v0
    request_buffering: true
    response_buffering: true
```

Elson9:
Fixes #822


src/lists/extensions/Namespace.ts
```diff
@@ -241,7 +241,7 @@ module.exports = {
               info: any,
               { query, access }: any
             ) => {
-              const namespaceValidationRule = '^[a-z][a-z0-9-]{4,14}$';
+              const namespaceValidationRule = '^[a-z][a-z0-9-]{3,13}[a-z0-9]$';
```
Jonesy, line 244: Won't block the review but if these are just same variable, hoist them up as a single global const so you only have to change it once next time. Cuts down memory too.

Jonesy (approved):

ikethecoder (approved):

PR #981, opened 2024-02-08
add support for protected externally

ikethecoder:
- Closes: #983 

Feature URL: https://api-services-portal-feature-protected-externally.apps.silver.devops.gov.bc.ca/


rustyjux:
A few more locations that look like they require adding a new option:
- `src\nextapp\components\environment-config\environment-config.tsx`
- `src\nextapp\components\environments-list\edit-environment.tsx`
- `src\nextapp\shared\services\utils.ts`, `getAuthToken()` around line 19
- `src\services\workflow\validate-active-environment.ts`

rustyjux (approved):
I think we've got it covered!

PR #1009, opened 2024-03-08
Remove Limit section in API directory for LOC

Elson9:
image

### DIFF
--- a/src/nextapp/components/api-product-item/api-product-item.tsx
+++ b/src/nextapp/components/api-product-item/api-product-item.tsx
@@ -56,7 +56,9 @@ const ApiProductItem: React.FC<ApiProductItemProps> = ({
                   as={isPublic || isTiered ? RiEarthFill : FaLock}
                   color="bc-blue"
                   boxSize="5"
-                  data-testid={`product-icon-${kebabCase(data.name)}-${isPublic || isTiered ? 'RiEarthFill' : 'FaLock'}`}
+                  data-testid={`product-icon-${kebabCase(data.name)}-${
+                    isPublic || isTiered ? 'RiEarthFill' : 'FaLock'
+                  }`}
                 />
               </Flex>
               <Heading size="xs">{data.name}</Heading>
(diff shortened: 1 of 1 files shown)


src/nextapp/components/api-product-item/api-product-item.tsx
```diff
@@ -56,7 +56,9 @@ const ApiProductItem: React.FC<ApiProductItemProps> = ({
                   as={isPublic || isTiered ? RiEarthFill : FaLock}
                   color="bc-blue"
                   boxSize="5"
-                  data-testid={`product-icon-${kebabCase(data.name)}-${isPublic || isTiered ? 'RiEarthFill' : 'FaLock'}`}
+                  data-testid={`product-icon-${kebabCase(data.name)}-${
```
Elson9, line 59: Just formatting here

rustyjux (approved):
I like that the LOC specific changes are isolated to single, removable lines. I also like that the special case for LOC is just to use the !isTiered && isGatewayProtected style, rather than a new condition.

But, like we discussed, it's very possible LOC will change the product name and then we are updating a component (and need to push a release) when a client specification changes.

I think the ideal scenario would be to give client control - e.g. add an optional field on Product which they can use to decide if they want the tiered description shown.

Another less intensive option might be to reference an external config file where we list exceptions. Perhaps more appropriate as part of `api-serv-infra`.

e2e/cypress/tests/09-update-product-env/09-two-tiered-hidden.cy.ts
```diff
@@ -0,0 +1,145 @@
...
+        nameSpace = twoTieredHidden.namespace
+        home.useNamespace(twoTieredHidden.namespace)
+        cy.get('@login').then(function (xhr: any) {
+          userSession = xhr.response.headers['x-auth-request-access-token']
+        })
+      })
+    })
+  })
+
+  it('update the Dataset in BC Data Catalogue to appear the API in the Directory', () => {
```
rustyjux, line 80:
```suggestion
  it('set the Dataset for the Product', () => {
```

src/nextapp/components/api-product-item/api-product-item.tsx
```diff
@@ -68,22 +76,32 @@ const ApiProductItem: React.FC<ApiProductItemProps> = ({
...
           </GridItem>
         </Grid>
-        {!isTiered && isGatewayProtected && (
+        {(!isTiered && isGatewayProtected) || isTieredHidden ? (
           <AccessRequestForm
             disabled={false}
             id={id}
             name={data.name}
             preview={preview}
+            dataTestId={
```
rustyjux, line 85: I'm not sure if dataTestId is the best name for this attribute - all these other props could be / are used to define the appearance of the AccessRequestDialog based on the context, so I'd stay stick with that convention (use a name which speaks to the data or desired behaviour).

How about just passing isTieredHidden as a prop and then using that as the conditional to determine the `data-testid` in <Button> or elsewhere as needed, rather than determining the final `data-testid` here in advance?

rustyjux (changes requested):
See comment

rustyjux (approved):

ikethecoder (approved):

PR #1010, opened 2024-03-08
Limits changed to Elevated Access, margin added + copy changed

Elson9:
image

Margin added to prevent two-tiered access from looking like a separate product.
Removed "has a rate limit enforced" as this is not a strict requirement of two-tiered access.
Limits changed to Elevated Access to avoid confusion.
Added text to clarify that public access does not require an API key.

---
🚀 Feature branch deployment: https://api-services-portal-feature-directory-changes.apps.silver.devops.gov.bc.ca

### DIFF
--- a/src/nextapp/components/api-product-item/api-product-item.tsx
+++ b/src/nextapp/components/api-product-item/api-product-item.tsx
@@ -81,16 +81,16 @@ const ApiProductItem: React.FC<ApiProductItemProps> = ({
           <Button
             disabled={true}
             data-testid="no-api-key-button"
-            variant={'primary'}
+            variant={'secondary'}
             fontWeight="600"
-            color="white"
+            color="bc-blue"
           >
-            No API Key Required
+            No API key required
           </Button>
         )}
       </Flex>
       {isTiered && (
-        <Flex px={9} py={7} bg={'white'} my={-5} mb={'0.5'}>
+        <Flex px={9} py={7} bg={'white'} my={-8} mb={'0.5'}>
           <Grid gap={4} flex={1} templateRows="auto" ml={8} mr={12}>
             <GridItem>
               <Flex align="center" mt={3}>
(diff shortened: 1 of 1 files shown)


rustyjux:
Design wise I think it's a step in the right direction towards clarifying that elevated access is not a separate product. 

Additional changes to consider:
- Reduce top margin (above Elevated Access).
- "Public access does not require..." could be included directly under the public product name which it relates to.
- Consider removing "For elevated access, please request access" entirely, since it should be more apparent what the "Request Access" button relates to.

Elson9:
Add in placeholder button

Elson9:
Latest changes:
image

rustyjux:
Hey nice. Three thoughts:
- Instead of disabled could we see the secondary style? (to compare at least)
- Further reduce padding/margins - there is 40 px between the public/elevated buttons right now, while there is only 30 px between the buttons and the dividers beyond. I feel like these should be same value, or if anything the space between the associated items should actually be less.
- Use sentence case 'No API key required' - I think the title case (Request Access) is appropriate when we have buttons with short names but otherwise sentence case is consistent with elsewhere (e.g. "View metrics in real-time") and looks better (IMO)

Elson9:
Thanks for your suggestions. Made those changes:
image

src/nextapp/components/api-product-item/api-product-item.tsx
```diff
@@ -68,31 +70,35 @@ const ApiProductItem: React.FC<ApiProductItemProps> = ({
...
+        {!isTiered && isGatewayProtected ? (
           <AccessRequestForm
             disabled={false}
             id={id}
             name={data.name}
             preview={preview}
           />
+        ) : (
+          <Button
+            disabled={true}
```
rustyjux, line 82:
```suggestion
```

rustyjux: or rather `={false}` if we keep that attribute around usually

Issue #980, opened 2024-02-07
Consumer app and app owner metadata


ikethecoder:
This ticket is for work to add the Application and Application Owner as metadata to the Consumer record so that it is available in the enhanced analytics.

ikethecoder:
The change involves:
- Adding a parameter for `tags` to the function https://github.com/bcgov/api-services-portal/blob/dev/src/services/kong/consumer-service.ts#L60 and using those values to be populated for the Consumer
- Update the callers of this `createKongConsumer` function to prepare the tags using the corresponding application and application owner, where the tags will have the format: `[ ``app:${app.name}``, ``owner:${app.owner.name}`` ]`

Elson9:
Draft pull request created: #1011

Elson9:
Fixed in #1028

PR #1011, opened 2024-03-11
Added Application and Application Owner as metadata to the Consumer r…

Elson9:
…ecord

---
🚀 Feature branch deployment: https://api-services-portal-feature-app-metadata.apps.silver.devops.gov.bc.ca

### DIFF
--- a/src/services/kong/consumer-service.ts
+++ b/src/services/kong/consumer-service.ts
@@ -10,6 +10,7 @@ import {
   KeyAuthResponse,
   KongConsumer,
 } from './types';
+import { Application } from '../keystone/types';
 
 const logger = Logger('kong.consumer');
 
@@ -42,7 +43,7 @@ export class KongConsumerService {
   public async createOrGetConsumer(
     username: string,
     customId: string,
-    app: any
+    app: Application
   ): Promise<CreateOrGetConsumerResult> {
     logger.debug('createOrGetConsumer');
     try {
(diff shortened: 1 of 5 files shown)


Elson9:
Fixes #980

src/services/kong/consumer-service.ts
```diff
@@ -51,16 +52,20 @@ export class KongConsumerService {
       return { created: false, consumer: result };
     } catch (err) {
       logger.debug('createOrGetConsumer - CATCH ERROR %s', err);
-      const result = await this.createKongConsumer(username, customId);
+      const result = await this.createKongConsumer(username, customId, app);
       logger.debug('createOrGetConsumer - CATCH RESULT %j', result);
       return { created: false, consumer: result };
```
Elson9, line 57: Should `created: true` here?

ikethecoder: yes think so

src/services/kong/consumer-service.ts
```diff
@@ -51,16 +52,20 @@ export class KongConsumerService {
...
     }
   }
 
-  public async createKongConsumer(username: string, customId: string) {
+  public async createKongConsumer(
+    username: string,
+    customId: string,
+    app: any
```
ikethecoder, line 64: Instead of "any", should be able to use "Application"

Elson9: Fixed

src/services/kong/consumer-service.ts
```diff
@@ -51,16 +52,20 @@ export class KongConsumerService {
...
+  ) {
     let body: KongConsumer = {
       username: username,
-      tags: ['aps-portal'],
+      tags: ['aps-portal', `app:${app.name}`, `owner:${app.owner.name}`],
```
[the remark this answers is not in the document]
Elson9: Since owner is part of the app object, wouldn't it only exist if app isn't null?

rustyjux (approved):

ikethecoder (approved):

PR #1016, opened 2024-03-22
Dataset formatting fix

ikethecoder:
New rendering of markdown on the API Directory detail page:

image

---
🚀 Feature branch deployment: https://api-services-portal-feature-dataset-markdown.apps.silver.devops.gov.bc.ca

### DIFF
--- a/src/nextapp/shared/styles/markdown.module.css
+++ b/src/nextapp/shared/styles/markdown.module.css
@@ -1,6 +1,22 @@
 .markdown {
 }
 
+.markdown h1 {
+  margin-bottom: 1em;
+}
+
+.markdown h2 {
+  margin-bottom: 1em;
+}
+
+.markdown h3 {
+  margin-bottom: 1em;
+}
+
+.markdown h4 {
+  margin-bottom: 1em;
+}
+
 .markdown ul {
   margin-top: 1em;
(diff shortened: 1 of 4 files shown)


src/nextapp/shared/styles/markdown.module.css
```diff
@@ -0,0 +1,72 @@
...
+}
+
+.markdown ol li {
+  margin-left: 2em;
+  display: list-item;
+  text-align: -webkit-match-parent;
+}
+
+.markdown img {
+  display: none;
```
rustyjux, line 44: This is intentional?
Not allowing images seems reasonable.

rustyjux (approved):
Looks good. 
In addition to the markdown sample dataset, I viewed the BC Route Planner item through the mock and I assume the content for this item was derived from the BCDC and so we shouldn't be creating any new problems.

rustyjux:
Also wondering about updating mock server docs - https://github.com/bcgov/api-services-portal?tab=readme-ov-file#mock-server
Seems to be as simple as running `npm run nextapp-dev` (making sure nothing else is running in docker). Are the other bits still relevant or accurate?
- Next and GWA API_ROOT
- Updating mock server schemas

Elson9 (approved):

PR #1022, opened 2024-03-28
Cypress astra local

nirajCITZ:
Added Astra call to APS API test scripts and updated docker compose file for keycloak image


ikethecoder (dismissed):

rustyjux:
Q: if Astra picks up vulnerabilities of an unacceptable impact level (currently high or medium), how can we view the scan results with the vulnerability details?

It seems to me the results are only written to `cypress/fixtures/state/scanResult.json` and we need to add another upload step to `.github/workflows/aps-cypress-e2e.yaml` so we can access the results.

rustyjux:
@ikethecoder @Elson9 this is ready for review now.

I added a Python script to create a Jira issue with the relevant vulnerability details.

I also added `exit 1` on the step which reviews the Cypress results to cause a failed GHA status. If there are any high/med issues from Astra, there will be a failed test in Cypress > GHA will fail. And thus I added `if: failure()` to only run the Astra results processing steps in that case.

Example of when Astra hits vulnerability (I modified the severity up from Low - these aren't real issues)
image

[Example Jira issue](https://dpdd.atlassian.net/browse/APS-2467) (will delete later, just left it for review)

Elson9 (approved):

rustyjux (approved):

PR #1028, opened 2024-04-03
Fix unable to generate keys error

Elson9:
---
🚀 Feature branch deployment: https://api-services-portal-feature-app-metadata.apps.silver.devops.gov.bc.ca

### DIFF
--- a/src/services/kong/consumer-service.ts
+++ b/src/services/kong/consumer-service.ts
@@ -11,6 +11,7 @@ import {
   KongConsumer,
 } from './types';
 import { Application } from '../keystone/types';
+import { alphanumericNoSpaces } from '../utils';
 
 const logger = Logger('kong.consumer');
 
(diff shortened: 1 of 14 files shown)


rustyjux (approved):
I successfully created a credential (w/ Kong API key) on the feature deployment. 

I am curious to see the consumer details being passed through to Grafana but not sure if that works for feature deployments (since I was logging into the Portal w/ janis@idir).

src/services/kong/consumer-service.ts
```diff
@@ -51,20 +53,36 @@ export class KongConsumerService {
...
       username: username,
       tags: ['aps-portal'],
     };
     if (customId) {
       body['custom_id'] = customId;
     }
+    if (app) {
+      body.tags.push(
+        `app:${app.name
+          .replace(/[^A-Za-z0-9:-]/gim, '')
```
ikethecoder, line 77: I think this regexp replace should be pulled out into a function that you can have a unit test for, to confirm that it is going to produce the expected output.  And it will also avoid having the logic duplicated.

ikethecoder (changes requested):

ikethecoder (approved):

PR #1073, opened 2024-06-19
Feature/gateway breadcrumbs

Elson9:


ikethecoder (commented):
use-namespace-breadcrumbs and use-namespace-root-breadcrumbs seem very similar - can we not consolidate them into a single hook and parameterize any differences?

Elson9:
Will use `appendedBreadcrumbs` to check if it's the gateway details page. Will remove check for `displayName` as it's no longer required.

PR #1101, opened 2024-07-16
/devportal and /manager redirects

Elson9:


rustyjux (changes requested):
Looks good so far but the login link from the /manager/gateways (http://oauth2proxy.localtest.me:4180/login?f=%2Fmanager%2Fgateways in local test), leads to the **Login for Developers** with various login options. It should go to the API Provider login

Elson9:
> Looks good so far but the login link from the /manager/gateways (http://oauth2proxy.localtest.me:4180/login?f=%2Fmanager%2Fgateways in local test), leads to the **Login for Developers** with various login options. It should go to the API Provider login

Fixed in latest commit

rustyjux:
Still needs more work - the last commit adds `?identity=provider` to all restricted pages including `/devportal`, e.g. `/devportal/applications` which is not appropriate.
I will add a conditional to look at the route and add `provider` only when appropriate and otherwise use the default (developer).

PR #1107, opened 2024-07-17
Feature approve own request

ikethecoder:


rustyjux:
Even snazzier would be to not sure users w/o the right permissions the approve/reject buttons (or to change styling to disabled). But this will do for now.

rustyjux (approved):

PR #1120, opened 2024-08-07
Feature/portal api v3

Elson9:
Ran full cypress test suite before and after changes and observed no differences

---
🚀 Feature branch deployment: https://api-services-portal-feature-portal-api-v3.apps.silver.devops.gov.bc.ca


Elson9:
Closed - should merge into feature/ns-to-gw instead of dev

PR #1121, opened 2024-08-07
Portal uses API v3 where appropriate

### LINKED ISSUE #1076: Namespace to Gateway - Portal uses API v3 where appropriate
There are only a few REST API calls (rather than graphql) within the Portal (for the directory). These should be updated to use the new v3 api.
Search `api/v2` to find - be mindful some things will not change to v3 (like openapi spec for v2!).

Elson9:
Ran full cypress test suite before and after changes and observed no differences.

Fixes #1076


Elson9:
Fixes #1076

rustyjux (approved):
Looks like that's all of em!

PR #1123, opened 2024-08-09
return display name on /gateways

### LINKED ISSUE #1080: Namespace to gateway - get gateways api should return both ID and displayName
Currently https://api-gov-bc-ca.dev.api.gov.bc.ca/ds/api/v3/console/#/Gateways/gateway-list returns a `string[]`, but this should really return an array of `GatewayId` and `DisplayName`.

rustyjux:
- Closes #1080

### DIFF
--- a/src/controllers/v3/GatewayController.ts
+++ b/src/controllers/v3/GatewayController.ts
@@ -105,14 +105,19 @@ export class NamespaceController extends Controller {
    */
   @Get()
   @OperationId('gateway-list')
-  public async list(@Request() request: any): Promise<string[]> {
+  public async list(@Request() request: any): Promise<{ gatewayId: string, displayName: string }[]> {
     const result = await this.keystone.executeGraphQL({
       context: this.keystone.createContext(request),
       query: list,
     });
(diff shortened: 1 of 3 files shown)


src/controllers/v3/GatewayController.ts
```diff
@@ -99,20 +99,25 @@ export class NamespaceController extends Controller {
...
   /**
-   * @summary List of Gateway IDs
+   * @summary List of Gateways available to the user
    * @param request
    * @returns
    */
   @Get()
   @OperationId('gateway-list')
-  public async list(@Request() request: any): Promise<string[]> {
+  public async list(@Request() request: any): Promise<{ gatewayId: string, displayName: string }[]> {
```
ikethecoder, line 108: May want to pull `type NamespaceInfo = { gatewayId: string, displayName: string }` out of this method so that it can be used in this Promise response as well.

rustyjux: doh we already had the `Gateway` type available! I updated to use that.

rustyjux:
Added a simple test.

PR #1144, opened 2024-08-19
Fix gateway detail redirect

### LINKED ISSUE #1122: Clicking on gateway list entry doesn't always redirect
# API Services Portal Issue

## User Story

As an API Provider, I want to reliably select a Gateway and have it redirect to the detail page every time.

### ENV

- [ x] `DEV`
- [x ] `TEST`
- [ x] `PROD`

### TESTCASE

- Go to detail page
- Select a Gateway

### EXPECTED

- Rediects to detail page

### ACTUAL

- Sometimes doesn't redirect

rustyjux:
- Fixes #1122 

Apparently adding `return null` to the `useEffect()` fixes the component render cycle and may be preventing the component from mounting and unmounting unpredictably before.

Worked 10x in a row in `docker compose` where it was frequently failing before. 

---
🚀 Feature branch deployment: https://api-services-portal-feature-ns-to-gw-gateway-detail-redirect.apps.silver.devops.gov.bc.ca


rustyjux:
Why are so many old commits showing? When I run `git merge dev` I get `Already up to date`.

src/nextapp/pages/manager/gateways/detail.tsx
```diff
@@ -158,6 +158,9 @@ const NamespacesPage: React.FC = () => {
       text: 'Your Organization and Business Unit will appear here',
     };
   }, [namespace]);
+  useEffect(() => {
+    window.scrollTo(0, 0);
+  }, []);
```
rustyjux, line 163: This change is to avoid the /detail page loading scrolled down when navigated to from /list if the user has scrolled down there. Maintaining scroll is noticeable and problematic for users on smaller displays. 

We could add `namespace` as a dependency in order to also scroll to top when changing GW using the dropdown. However, this also causes scroll to top when going back from /manager page, which seems undesirable.

In newer versions of NextJS, the default option for router.push is scroll to top (and one can easily modify it for each call) but that's a much bigger can of worms.

PR #1179, opened 2024-09-13
clickable items in Gateways list

### LINKED ISSUE #1167: Enable clickable gateway list item panel for improved navigation
# API Services Portal Issue

## Story Details

**As an** API Provider,
**I want** the entire gateway list item panel to be clickable,
**so that** I can easily navigate to the gateway details page without confusion.

**Current Behavior:**
```
The user currently needs to select the gateway name text to view the gateway details page. This approach lacks clarity and does not provide a good user experience, as it is not evident that selecting the gateway name will display the details.
```

**Proposed Behavior:**
```
The user should be able to click anywhere within the gateway list item to navigate to its details page. This change will enhance clarity and improve the user experience by clearly indicating that selecting the entire panel will display the gateway details.
```

## Acceptance Criteria

- Clickable Area: The entire area of the gateway list item panel should be clickable, not just the gateway name text.
-  Navigation: Clicking anywhere within the gateway list item panel should navigate the user to the corresponding gateway details page.
-  Visual Indication: The list item panel should have a visual indication (e.g., hover effect, cursor change) to show that it is clickable.
-  Consistency: The new behavior should be consistent across all gateway list items and within all supported browsers and devices.
-  User Feedback: Ensure that users are able to easily find and access the gateway details page with minimal confusion.

rustyjux:
- fixes #1167 

I believe this is the first instance in which we have a clickable item / card sort of situation. This is a little one-off addition using Chakra inline styles but if we wanted to implement this sort of thing more widely, we could come up with a reusable component.

Or we could continue to use inline styles and handle each component individually.

---
🚀 Feature branch deployment: https://api-services-portal-feature-click-gateway-item.apps.silver.devops.gov.bc.ca

### DIFF
--- a/src/nextapp/pages/manager/gateways/list.tsx
+++ b/src/nextapp/pages/manager/gateways/list.tsx
@@ -17,7 +17,6 @@ import Head from 'next/head';
 import { gql } from 'graphql-request';
 import { FaPlus, FaLaptopCode, FaRocket, FaServer } from 'react-icons/fa';
 import { useQueryClient } from 'react-query';
-import { css } from '@emotion/react';
 
 import PageHeader from '@/components/page-header';
 import GridLayout from '@/layouts/grid';
@@ -188,14 +187,6 @@ const MyGatewaysPage: React.FC = () => {
     }
   }, [data, search, filter]);
 
-  const clickableItemStyle = css`
-    cursor: pointer;
-    transition: background-color 0.2s;
-    &:hover {
-      background-color: #f0f0f0;
-    }
-  `;
-
   return (
     <>
       <Head>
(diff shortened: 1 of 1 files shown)


timng-bcgov (approved):
I have two observations on the styling of the list items. 

1. I'm wondering if we're able to style the Gateway display name with an underline based on the state of the parent to better match the styling of cards in the Gateways manager:

Current:
image

Comparison (expected from Gateway details):
image

Quick mockup to illustrate:
image


2. Should we think about focus states for the list items? They seem to be absent on a few links throughout the Services portal. Is it possible to use a Chakra Link component (which composes `<Box />` and has the desired focus state) instead of a `<Flex />`  with an `onClick` event? Might be worth discussing the composition here if we are planning on making a reusable component.

Elson9 (approved):

PR #1184, opened 2024-10-04
tweak style for gateway list item hover

rustyjux:
Follows up on Tim's comment on #1179 - https://github.com/bcgov/api-services-portal/pull/1179#pullrequestreview-2307883242
> I'm wondering if we're able to style the Gateway display name with an underline based on the state of the parent to better match the styling of cards in the Gateways manager:

image

We think the hover background is rather dark but apparently that's the recommendation in the BC design tokens.

---
🚀 Feature branch deployment: https://api-services-portal-feature-click-gateway-item.apps.silver.devops.gov.bc.ca

---
🚀 Feature branch deployment: https://api-services-portal-feature-click-gateway-item.apps.silver.devops.gov.bc.ca

### DIFF
--- a/src/nextapp/pages/manager/gateways/detail.tsx
+++ b/src/nextapp/pages/manager/gateways/detail.tsx
@@ -247,7 +247,7 @@ const NamespacesPage: React.FC = () => {
                   <PopoverArrow bgColor="#373d3f" />
                   <PopoverBody>
                     If you need to change the Organization or Business Unit for
-                    your Namespace, submit a request through the{' '}
+                    your Gateway, submit a request through the{' '}
                     <Link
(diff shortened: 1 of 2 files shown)


timng-bcgov (approved):
For reference, I'm following guidelines for the `secondary` [button component](https://www2.gov.bc.ca/gov/content/digital/design-system/components/buttons#) from the Design System. The button doesn't toggle the label colour like we're doing here, but there are no concerns about contrast:
image

We still need to work on the keyboard focus state and semantic identification of the component (use a button or anchor element instead of a div).

rustyjux:
@Elson9 waiting for review

ikethecoder (approved):

Issue #289, opened 2022-01-14
API Provider - API Terms of Use


elisafw-ux:
Story: As an API administrator, I need an automated way to communicate my API terms of use with a user, allow the user to read and accept the terms before they are granted access to my API, and have the acceptance recorded as a record of decision for future auditing if required. (OR) As an API administrator I need to have an ability to broadcast updates to terms to all registered API users w/o having them to agree to the terms again

*Details:* There is a "Legal" list that is defined that can be set for a Product Environment, and is then displayed on the Request Access page.  Additional work is needed to update the User profile with this acceptance.

A/C:

Show an API Owner selecting the Terms of Use Legal Terms when defining an Environment

Show a Developer requesting access and accepting the Terms of Use

Show the Developer going in again and not seeing the Terms of Use because they have already agreed to them

ikethecoder:
Some work to make the Terms of Use details available on the User Profile page, and enhancement to the Directory API to allow API Providers a way of managing new Legal terms.

PR #1185, opened 2024-10-16
fix tests to handle redirect away from your products

rustyjux:
Fixes:
- #1181

### DIFF
--- a/.github/workflows/aps-cypress-e2e.yaml
+++ b/.github/workflows/aps-cypress-e2e.yaml
@@ -76,6 +76,12 @@ jobs:
           name: code-coverage
           path: ${{ github.workspace }}/e2e/coverage
 
+      - name: Upload Astra scan results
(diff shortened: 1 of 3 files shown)


Elson9 (approved):

rustyjux:
I'm not entirely sure why https://github.com/bcgov/api-services-portal/actions/runs/11373768072/job/31641057917 failed but I added the upload of the Astra scan result so if things do pop up there in future we can at least see the scan result.

rustyjux:
Also slipped in an update to fix broken link for 'View legal' on Request Access page so at least it is not broken.
This link isn't (and never was) entirely appropriate since it is regarding Open Government Licence but even APIs which are not providing open data link to the same terms. #289 elaborates a little more on possible enhancements here.

ikethecoder (approved):

PR #1217, opened 2025-01-07
fix for click on gateway error

### LINKED ISSUE #1122: Clicking on gateway list entry doesn't always redirect

ikethecoder:
- Fixes #1122 

---
🚀 Feature branch deployment: https://api-services-portal-feature-click-on-gateway.apps.silver.devops.gov.bc.ca

---
🚀 Feature branch deployment: https://api-services-portal-feature-click-on-gateway.apps.silver.devops.gov.bc.ca

### DIFF
--- a/src/nextapp/pages/devportal/api-directory/your-products.tsx
+++ b/src/nextapp/pages/devportal/api-directory/your-products.tsx
@@ -29,7 +29,7 @@ const ApiDiscoveryPage: React.FC = () => {
   const hasNamespace = !!user?.namespace;
   const title = (
     <>
-      {(namespace.isFetching || namespace.isLoading) && (
+      {(namespace.isFetching) && (
         <Skeleton width="400px" height="20px" mt={4} />
       )}
       {namespace.isSuccess && !namespace.isFetching && (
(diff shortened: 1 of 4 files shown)


Elson9 (approved):
Tested locally and working well with fixing the redirect issue. Also redirects to the list page when a gateway isn't selected (i.e. when trying to access the detail page on first login)

rustyjux:
Clean and simple. Works well on feature deployment and local, hopefully it holds true!

It seems like in instances where we have `namespace.isFetching || namespace.isLoading` as a condition, the `isLoading` part is redundant since `isFetching` includes the initial load (unlike isRefetching which is only refetches). I made a small commit to reflect this; tested locally and the skeleton still shows when expected.

rustyjux (approved):

PR #1227, opened 2025-02-01
Cypress kong plugins

ikethecoder:


local/kong/Dockerfile
```diff
@@ -1,50 +1,30 @@
...
-RUN luarocks install lua-resty-openidc
-RUN git clone -b v${PLUGIN_OIDC_VERSION} https://github.com/revomatico/kong-oidc.git
-RUN (cd kong-oidc && luarocks make && luarocks pack kong-oidc ${PLUGIN_OIDC_VERSION})
-
-RUN git clone https://github.com/ikethecoder/kong-oidc-consumer.git
-RUN (cd kong-oidc-consumer && luarocks make && luarocks pack kong-oidc-consumer ${PLUGIN_OIDC_CONSUMER_VERSION})
-
-RUN git clone -b kong28 https://github.com/ikethecoder/kong-plugin-jwt-keycloak.git
-RUN (cd kong-plugin-jwt-keycloak && luarocks make && luarocks pack kong-plugin-jwt-keycloak ${PLUGIN_VERSION})
+RUN git clone -b cleanup https://github.com/bcgov/kong-oss-plugins.git \
```
rustyjux, line 7: Will need to adjust the branch here once `cleanup` is merged in.

Do we need a way to control plugin versions individually or we should be fine to use `main`?

rustyjux (approved):

local/kong/Dockerfile
ikethecoder: After merging kong-oss-plugins into main, I created a tag `v1.5.0-1` - so updated to use that.

Elson9 (approved):

PR #1228, opened 2025-02-06
hotfix ops metrics

ikethecoder:


rustyjux (approved):
The new function seems fine. Funny we see this paging issue again - didn't this come up some other time in the last year? or was that another query?

I made a small change to fix an error during `npm run build` (see https://github.com/bcgov/api-services-portal/actions/runs/13190804790/job/36823280137 - Build Docker Images step).

I can approve, but I have a couple questions:
1. What issue is this trying to fix? 
2. How do I run the tests? are the more instructions on 'wiring up directly with keycloak'?

ikethecoder:
Before it was hard coded to 1000 records - now it will make multiple paging calls to get a full list.. so it is a preemptive change.. we probably should look at demising `listAll` in favour of `listAllPages` at some point to avoid a "mysterious missing records" problem.

The `test/integrated` folder is adhoc testing and so you have to be in a state where you can run the Portal locally, but instead of running the Portal locally, you run:

```sh
npm run ts-build
npm run ts-watch
```

..in one terminal and this in another:

```sh
node dist/test/integrated/batchworker/paging.js
```

Elson9 (approved):

PR #1311, opened 2025-12-17
pre defined patterns

ikethecoder:
---
🚀 Feature branch deployment: https://api-services-portal-feature-pre-defined-patterns.apps.silver.devops.gov.bc.ca

### DIFF
--- a/src/controllers/types.ts
+++ b/src/controllers/types.ts
@@ -1,0 +1,6 @@
+export class TsoaErrorWrapper {
+  inner: Error;
+  constructor(inner: Error) {
+    this.inner = inner;
+  }
+}
--- a/src/controllers/types.ts
+++ b/src/controllers/types.ts
@@ -1,6 +1,0 @@
-export class TsoaErrorWrapper {
-  inner: Error;
-  constructor(inner: Error) {
-    this.inner = inner;
-  }
-}
--- a/src/controllers/v3/GatewayConfigController.ts
+++ b/src/controllers/v3/GatewayConfigController.ts
@@ -14,12 +14,11 @@ import {
 import { KeystoneService } from '../ioc/keystoneInjector';
 import { inject, injectable } from 'tsyringe';
 import { GetConfigUsingPattern } from '../../services/gateway-patterns/evaluator';
-import { TsoaErrorWrapper } from './types-extra';
+import { TsoaErrorWrapper } from '../types';
 
 interface GatewayPatternConfigRequest {
   pattern: string;
-  delete?: boolean;
-  parameters: any;
+  parameters: { [key: string]: string };
 }
 
 @injectable()
--- a/src/controllers/v3/GatewayConfigController.ts
+++ b/src/controllers/v3/GatewayConfigController.ts
@@ -9,10 +9,12 @@ import {
   Security,
   Body,
   Tags,
+  Example,
 } from 'tsoa';
 import { KeystoneService } from '../ioc/keystoneInjector';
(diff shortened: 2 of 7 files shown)


src/controllers/types.ts
```diff
@@ -0,0 +1,6 @@
+export class TsoaErrorWrapper {
```
phowells: Best practice is to extend Error and capture the cause in the options.  Something like this.

new TsoaErrorWrapper("Some message for additional context", { cause: originalError });

export class TsoaErrorWrapper extends Error {
  constructor(message, options) {
    super(message, options);
  }
}

src/controllers/v3/GatewayConfigController.ts
```diff
@@ -0,0 +1,60 @@
...
+  @OperationId('get-config-using-pattern')
+  @Security('jwt', [])
+  @Example<GatewayPatternConfigRequest>({
+    pattern: 'simple-service.r1',
+    parameters: {
+      service_name: 'my-service',
+      service_url: 'https://httpbun.com',
+    },
+  })
+  public async put(
```
phowells: Should the function name align with the operation Id rather than the HTTP method?

src/controllers/v3/GatewayConfigController.ts
```diff
@@ -0,0 +1,60 @@
...
+  ): Promise<any> {
+    // always inject the gatewayId as a parameter
+    body.parameters.gateway_id = gatewayId;
+
+    const ctx = this.keystone.createContext(request);
+
+    try {
+      return await GetConfigUsingPattern(ctx, body);
+    } catch (error) {
+      throw new TsoaErrorWrapper(error);
```
phowells: It would be good if we could return some more info with the errors.  If this is a validation error on the input then we should return a 400 with details.  I think this will result in a 500 error.  I think tsoa defines a ValidationError class for this.

phowells (commented):
Not sure if this is the kind of feedback you are looking for.

src/controllers/v3/GatewayConfigController.ts
rustyjux: it does return a 400 with details, the logic is in `src/services/gateway-patterns/evaluator.ts` and `src/api-openapi.js`

rustyjux (approved):
Looks good. Can't speak to Paul's comments except to note that the error messaging seems fine (400s for missing parameters or invalid pattern, 422 if `pattern` or `parameters` is missing, with descriptive messages all around).

Could add a test for unsupported pattern.

src/controllers/v3/GatewayConfigController.ts
phowells: Sorry, I am just getting familiarizing with this framework.  Looks like we are using 422 for validation errors and some syntax errors but with formats.  This is existing code.  The errors being wrapped by TsoaErrorWrapper look like validation errors on the input to me.  If they are then could we be using ValidationError and providing field level details for the errors.

ikethecoder:
Thanks @phowells and @rustyjux  for feedback - I did some refactoring, particularly on the errors.  Have another review and let me know if any further changes are needed.

phowells (approved):
Looks good.  Thank you.

Elson9 (approved):

rustyjux:
Just occurred to me a helpful addition (which could come down the road) would be an endpoint returning available patterns, their parameters, and potentially also a short description or example pattern input. This would allow for showing this info in a UI or CLI without needing to hardcode it in each frontend.

PR #1313, opened 2025-12-24
SDX subsystems, services, runtime groups and connections

ikethecoder:
Closes the following issues:

- APS-3961 Discover my Service on SDX
- APS-3966 Assign SDX Edge Server to Gateway 
- APS-4047 Grant access - client and service
- APS-4050 Provision new Runtime Group
- APS-4054 Apply a minimal SDX specific pattern
- APS-4060 Subsystems and Services
- APS-4214 Provision default routes for edge servers
- APS-4215 Ensure data integrity between consumer & provider

---
🚀 Feature branch deployment: https://api-services-portal-feature-sdx-service.apps.silver.devops.gov.bc.ca


e2e/cypress/support/prep-commands.ts
```diff
@@ -45,7 +45,7 @@ Cypress.Commands.add('buildOrgGatewayDatasetAndProduct', (): Cypress.Chainable<a
               member: {
                 email: 'janis@testmail.com',
               },
-              roles: ['organization-admin'],
+              roles: ['organization-admin', 'system-owner'],
```
rustyjux, line 48: could also make Benny a system owner (in addition to org admin) on `ca.bc.gov` in https://github.com/bcgov/api-services-portal/blob/feature/sdx-service/local/keycloak/master-realm.json#L900 to facilitate interactive tests without running these prep commands

src/controllers/sdx/v1/OrgServiceController.ts
```diff
@@ -0,0 +1,189 @@
...
+    const batchClause = {
+      query: '$org: String',
+      clause: '{ organization: { name: $org } }',
+      variables: { org },
+    };
+    return await GetCatalog(ctx, false, batchClause);
+  }
+
+  /**
+   * Retrieve an oas-services associated with an organization
```
rustyjux:
```suggestion
   * Retrieve an oas-service associated with an organization
```

src/authz/matrix.csv
rustyjux, line 1: it would be great if this diff was accurate. Hard to know if it's just those few Org Admin lines up top or more throughout.

rustyjux: could also update the readme w/ new lists
https://github.com/bcgov/api-services-portal/blob/7ff981685ce86e3f4be3f33ffdd29294179aa3c5/src/authz/README.md

src/services/batch/subsystem.ts
```diff
@@ -0,0 +1,91 @@
...
+import { regExprValidation } from '../utils';
+import { strict as assert } from 'assert';
+import { Subsystem } from './types';
+import { Subsystem as KeystoneSubsystem } from '../keystone/types';
+import { Keystone } from '@keystonejs/keystone';
+
+class SubsystemService {
+  validateSubsystem = (name: string): void => {
+    regExprValidation(
+      '^[A-Z0-9-]{3,20}$',
```
rustyjux, line 24: Should this be capitals only? if so, validation error should reflect this

rustyjux (changes requested):
Wow there's a lot in here!

Generally looks good. A few questions: 

1. Seems like subsystem and oas-service are just there for cataloging; will oas-service be linked up to actual (SDX) services at some stage? oas-service be a child of an SDX service?
2. For catalog/discovery purposes, do we need a way of linking users/contacts to oas-services (or perhaps subsystems)? Or are we good with oas-service > subsystem > gateway > owners / access managers, which should be accessible with the current path?
3. Forget about org units? too fickle?

I see mocks haven't been created for the new endpoints but we can add those later if they are deemed necessary for future frontend dev.

**Request for changes:** A few new test failures have arisen which need addressing - https://github.com/bcgov/api-services-portal/issues/1328. Only these 3 are expected/acceptable w/ Kong 3 - https://github.com/bcgov/api-services-portal/issues/1306

I also left one or two comments of significance.

phowells (approved):
A lot to look at here.  It would be helpful if there were more in-line comments on the code as well as in-line comments in the PR explaining why you made changes.

Elson9 (approved):

rustyjux (approved):

PR #1331, opened 2026-01-09
APS-3966, APS-4050 - Runtime Groups

ikethecoder:
Resolves:
- APS-4050 Provision new Runtime Group
- APS-3966 Assign SDX Edge Server to Gateway

Available in DEV: https://api-gov-bc-ca.dev.api.gov.bc.ca/

Relevant docs [internal]: https://github.com/bcgov-c/api-serv-infra/blob/docs/ops/docs/playbooks/SDX.md#sdx-edge-server-deployment


rustyjux:
It would be nice to see new e2e tests pass in GH and to know the required gwa-api branch. Can you update https://github.com/bcgov/api-services-portal/blob/feature/runtime-groups/.github/workflows/aps-cypress-e2e.yaml#L26 and kick off a workflow run?

Also could use entries in https://github.com/bcgov/api-services-portal/blob/feature/runtime-groups/e2e/test-data-dependencies.md

src/lists/RuntimeGroup.js
```diff
@@ -0,0 +1,54 @@
...
+      isUnique: true,
+    },
+    hostedOrganizations: {
+      type: Relationship,
+      ref: 'Organization',
+      many: true,
+    },
+    sdxEndpoint: {
+      type: Text,
+      isRequired: false,
```
rustyjux: why are `sdxEndpoint` and `consumerEndpoint` not required? is it helpful to be able to create an rg without these defined and they can be updated later?

If they can be null, there might be an issue in CreateNamespaceForRuntimeGroup (workflow) since it calls this in a way that will throw an error:
https://github.com/bcgov/api-services-portal/blob/feature/runtime-groups/src/services/workflow/create-namespace-sdx.ts#L43 and again at 98

src/controllers/sdx/v1/OrgRuntimeGroupController.ts
```diff
@@ -0,0 +1,208 @@
...
+import { BatchResult } from '../../../batch/types';
+import {
+  deleteRecordByInternalId,
+  getRecords,
+  removeEmpty,
+  removeKeys,
+  replaceKey,
+  syncRecordsThrowErrors,
+  transformAllRefID,
+} from '../../../batch/feed-worker';
```
rustyjux: all these imports for feed-worker are unused

src/controllers/sdx/v1/OrgRuntimeGroupController.ts
```diff
@@ -0,0 +1,208 @@
...
+   * Constructor initializes the controller with Keystone service for context creation.
+   * @param _keystone - KeystoneService instance injected via dependency injection
+   */
+  constructor(@inject('KeystoneService') private _keystone: KeystoneService) {
+    super();
+    this.keystone = _keystone;
+  }
+
+  /**
+   * Create a new runtime group for an organization
```
rustyjux:
```suggestion
   * @summary Create or update a runtime group
```

src/controllers/sdx/v1/OrgRuntimeGroupController.ts
```diff
@@ -0,0 +1,208 @@
...
+    input['organization'] = org;
+
+    const service = new RuntimeGroupService();
+
+    // Create or update the runtime group
+    return service.upsertRuntimeGroup(ctx, input);
+  }
+
+  /**
+   * Retrieve the list of runtime groups associated with an organization
```
rustyjux:
```suggestion
   * @summary Retrieve a list of runtime groups
```

src/controllers/sdx/v1/OrgRuntimeGroupController.ts
```diff
@@ -0,0 +1,208 @@
...
+    // Return filtered runtime groups based on the specified filter
+    if (filter === 'available') {
+      return service.listHostedRuntimeGroupsForOrganization(ctx, org);
+    } else {
+      return service.listRuntimeGroupsByOrganization(ctx, org);
+    }
+  }
+
+  /**
+   * Delete a runtime group
```
rustyjux:
```suggestion
   * @summary Delete a runtime group
```

src/controllers/sdx/v1/OrgRuntimeGroupController.ts
```diff
@@ -0,0 +1,208 @@
...
+    const context = this.keystone.createContext(request, true);
+
+    const service = new RuntimeGroupService();
+
+    // Delete the runtime group (with force option if specified)
+    return service.deleteRuntimeGroup(context, org, name, force);
+  }
+
+  /**
+   * Register a gateway for a runtime group
```
rustyjux:
```suggestion
   * @summary Register a gateway for a runtime group
```

src/services/batch/runtime-group.ts
```diff
@@ -0,0 +1,175 @@
...
+    );
+  };
+
+  upsertRuntimeGroup = async (
+    context: Keystone,
+    body: RuntimeGroup
+  ): Promise<BatchResult> => {
+    // host should be based on a standard format for edge servers
+    body['host'] = `${body['name']}.servers.sdx`;
+    if (!body.hasOwnProperty('sdxEndpoint')) {
```
rustyjux:
```suggestion
    if (!body.sdxEndpoint) {
```
`hasOwnProperty` returns true if `null`, `undefined`, or `""` is passed, which would be problematic
Same for `consumerEndpoint`. just below

src/controllers/sdx/v1/OrgSubsystemController.ts
rustyjux, line 1: at line 62, should be `* Retrieve the list of subsystems associated with an organization`

src/controllers/sdx/v1/OrgSubsystemController.ts
```diff
@@ -65,6 +73,23 @@ export class OrgSubsystemController extends Controller {
     return new SubsystemService().listSubsystemsByOrganization(ctx, org);
   }
 
+  @Get('/{name}/client')
```
rustyjux: add API doc comment - what does `client` mean in this context?

src/controllers/sdx/v1/OrgSubsystemController.ts
```diff
@@ -88,4 +113,32 @@ export class OrgSubsystemController extends Controller {
 
     return new SubsystemService().deleteSubsystem(context, org, name, force);
   }
+
+  @Put('/{name}/gateway')
```
rustyjux, line 154: can add API doc comment 

more generally, wondering when this occurs vs registering gateway w/ runtime group. I don't see this endpoint in the docs

rustyjux (approved):
Looks good. I'd be happy to try walking through the process now or later if you think that'd be helpful. Permissions need to be manually assigned in Keycloak right?

More of a general Q: is there only one subsystem per runtime group? That seems to be the current structure, since they both have (the same) single gateway.

src/test/services/batch/testdata.js
```diff
@@ -210,7 +210,7 @@ export default {
       data: {
         namespace: 'refactortime',
       },
-      expected: { exception: 'Missing value for key name' },
+      expected: { exception: 'Missing value for key name\n\nfalse !== true\n' },
```
rustyjux: Do you understand where the change was introduced to cause error messages to include this stuff? Seems really not desirable for users

src/controllers/sdx/v1/OrgRuntimeGroupController.ts
ikethecoder: cleaned all up under sdx/v1

src/test/services/batch/testdata.js
ikethecoder: I introduced user-assert.ts which will do a comparison without adding the extra debug bits.  We will have to at some point refactor the other code to use it.

This: `import { strict as assert } from 'assert';`

To this: `import assert from '../user-assert';`

PR #1335, opened 2026-02-04
APS-4156 remove services from response

rustyjux:
## Summary

This PR sanitizes the Directory API response by removing the `services` payload and extracts the `twoTieredHidden` flag from plugin tags before services are removed. The UI is updated to use the extracted flag instead of accessing services directly.

The behaviour for two-tiered-hidden products is changed to show a lock icon instead of an earth (implying public), thereby _actually_ hiding two-tiered access. I couldn't find any tickets documenting the initial request or requirements from LOC, but noone is currently using this feature and I feel this is a more logical approach for potential future users.

### API Controllers
- **Remove services from directory API responses** to reduce payload size
- **Extract `twoTieredHidden` flag** from plugin tags (`aps.two-tiered-hidden`) before removing services
- Apply changes to:
  - v2 and v3 Directory and GatewayDirectory/NamespaceDirectory controllers
  - v1 not updated: no longer available

### Frontend
- Update `api-product-item` component to use `twoTieredHidden` property instead of accessing services
- Show lock icon (FaLock) for two-tiered-hidden products instead of public icon (RiEarthFill)

## Other Updates
- Local Kong version bump (3.9.0 → 3.9.1) and added curl to Dockerfile (fix build issue)
- README: Added WSL hostip command for OAuth2 proxy setup
- Updated e2e tests to expect lock icon for two-tiered-hidden products

---
🚀 Feature branch deployment: https://api-services-portal-feature-APS-4156-remove-services-from-response.apps.silver.devops.gov.bc.ca

---
🚀 Feature branch deployment: https://api-services-portal-feature-APS-4156-remove-services-from-response.apps.silver.devops.gov.bc.ca

PR #1339, opened 2026-02-05
APS-4156 remove services from directory response

### LINKED ISSUE #1334: Remove services from directory API response
# API Services Portal Bug Issue

`directory/{id}` returns services. It shouldn't.

removing services may break two-tiered-hidden, so keep an eye on that test case - see https://github.com/bcgov/api-services-portal/blob/feature/runtime-groups/src/nextapp/components/api-product-item/api-product-item.tsx#L48-L51

If so, might have to move some of the frontend logic into the directory API to just return the information it needs to render the UI (e.g. `aps` tags), rather than all the service/plugin info

### ENV

- [X] `DEV`
- [X] `TEST`
- [X] `PROD`

rustyjux:
Note: supercedes https://github.com/bcgov/api-services-portal/pull/1335 - branch name was too long, causing helm issues

## Summary

This PR sanitizes the Directory API response by removing the `services` payload and extracts the `twoTieredHidden` flag from plugin tags before services are removed. The UI is updated to use the extracted flag instead of accessing services directly.

The behaviour for two-tiered-hidden products is changed to show a lock icon instead of an earth (implying public), thereby _actually_ hiding two-tiered access. I couldn't find any tickets documenting the initial request or requirements from LOC, but noone is currently using this feature and I feel this is a more logical approach for potential future users.

Fixes #1334 

### API Controllers
- **Remove services from directory API responses** to reduce payload size
- **Extract `twoTieredHidden` flag** from plugin tags (`aps.two-tiered-hidden`) before removing services
- Apply changes to:
  - v2 and v3 Directory and GatewayDirectory/NamespaceDirectory controllers
  - v1 not updated: no longer available

### Frontend
- Update `api-product-item` component to use `twoTieredHidden` property instead of accessing services
- Show lock icon (FaLock) for two-tiered-hidden products instead of public icon (RiEarthFill)

## Other Updates
- Fixed minor validation error regressions introduced in https://github.com/bcgov/api-services-portal/pull/1311
- Local Kong version bump (3.9.0 → 3.9.1) and added curl to Dockerfile (fix build issue)
- README: Added WSL hostip command for OAuth2 proxy setup
- Updated e2e tests to expect lock icon for two-tiered-hidden products

---
🚀 Feature branch deployment: https://api-services-portal-feature-remove-services.apps.silver.devops.gov.bc.ca

---
🚀 Feature branch deployment: https://api-services-portal-feature-remove-services.apps.silver.devops.gov.bc.ca

### DIFF
--- a/.github/workflows/ci-feat-sonar.yaml
+++ b/.github/workflows/ci-feat-sonar.yaml
@@ -21,7 +21,7 @@ jobs:
 
       - uses: actions/setup-node@v2
         with:
-          node-version: '14'
+          node-version: '22'
 
       - name: Run Tests
         run: |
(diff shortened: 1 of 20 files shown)


Elson9 (approved):
Looks good to me.

danebenal (approved):

PR #1354, opened 2026-02-20
APS-4068 - use Postgres 15 instead of Mongo for feature deployments

rustyjux:
Feature deploymentss now use Postgres 15 (public `postgres:15` image) instead of MongoDB. Follows the same pattern as e2e (`ADAPTER=knex`, `local/db/keystone-init.sql`).

Postgres is deployed via inline k8s manifests (Deployment, Service, PVC, init ConfigMap, Keystone-schema Job). There is potential to use a helm chart but this seemed like the least overhead. Updated delete deployment (`ci-remove.yaml`) to handle this change.

Validation:
- normal Portal functionality (can update Keystone data)
- data is persisted across users/sessions, redeployments
- manually validated the delete deployment commands (prob need to merge into default branch `dev` to run in GH)

---
🚀 Feature branch deployment: https://api-services-portal-feature-feature-pg.apps.silver.devops.gov.bc.ca

### DIFF
--- a/.github/workflows/ci-build-deploy.yaml
+++ b/.github/workflows/ci-build-deploy.yaml
@@ -88,7 +88,7 @@ jobs:
           curl -L -O https://get.helm.sh/helm-v3.4.2-linux-amd64.tar.gz
           tar -xf helm-v3.4.2-linux-amd64.tar.gz
 
-      - name: 'Deploy Database (Postgres 15)'
+      - name: 'Deploy Database'
         if: github.ref != 'refs/heads/dev'
         run: |
           export PATH=$PATH:`pwd`/linux-amd64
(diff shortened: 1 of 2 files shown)


Elson9 (commented):
Looks good overall, my only concern is with storage-quota if a PVC is being created for each feature branch

rustyjux:
> Looks good overall, my only concern is with storage-quota if a PVC is being created for each feature branch

This was the existing pattern. I guess they could all share one PVC but then we hit issues if there are many feature branches and cleanup is trickier.

Probably more problematic than PVC per feature branch is that right now old feature deployments aren't being cleaned up (either branches aren't being deleted, or the cleanup failed - I haven't dug into why there are so many). I would like to verify the cleanup workflow works but it needs to get merged into `dev` to see it run.

In any case, we can definitely reduce the PVC size. Ever more long-lived feature branches seeing more activity like Aidan's `sdx` ones are only using 0.5 GiB. So rather than 2 GiB I'll drop it to 1.

danebenal (approved):

Elson9 (approved):

PR #1362, opened 2026-02-25
Add NPM_TOKEN

Elson9:
Fixes 403 Forbidden on Cypress tests.

The 403 Forbidden happens when the npm registry rate-limits unauthenticated requests. In our workflow, several images run npm install in parallel (cypress-jwks-url, feeder, apsportal, cypress, etc.), so the runner hits that limit and npm returns 403.


danebenal:
Do we have documentation detailing the need for the NPM_TOKEN and it's 90 days rotation somewhere?

Elson9:
> Do we have documentation detailing the need for the NPM_TOKEN and it's 90 days rotation somewhere?

Not yet. This is a newly discovered issue.

danebenal (approved):

rustyjux (approved):
Good catch. Thanks for laying it out clearly in the chat message.

PR #1369, opened 2026-03-10
Feature/kc 26

ikethecoder:
### Keycloak 15 → 26 upgrade support

- **Keycloak service layer**
  - Updated `client-service`, `client-policy-service`, `client-registration-service`, `client-roles`, `group-service`, `user-service`, `namespace-details`, `token-service`, and related org-group/UMA services to align with Keycloak 26 APIs and behavior.

- **Realm and local Keycloak setup**
  - Updated `local/keycloak/master-realm.json` for Keycloak 26 schema/fields.
  - Added `local/keycloak/Dockerfile` for Keycloak 26 and `local/keycloak/nginx-lb.conf` for the new deployment topology.

- **OAuth2-proxy and auth integration**
  - Adjusted `src/auth/auth-oauth2-proxy.js`, `src/auth/auth-tsoa.ts`, and added `src/auth/forbidden-error.ts` for Keycloak 26 token handling and forbidden responses.
  - Updated OAuth2-proxy configs under `local/oauth2-proxy/`.

- **E2E and tests for Keycloak 26**
  - Updated Cypress page objects and tests (Keycloak clients, scopes, groups, users, org assignment, APS API, permissions, shared IDP).
  - Updated fixtures and mocks (`src/test/mocks/handlers/data/keycloak.yaml`, `src/test/mocks/handlers/keycloak.ts`) and integrated Keycloak/org-groups tests.

### Other changes

- **CI and test tooling**
  - Tweaks to GitHub Actions workflows, Jest/babel setup, and batch worker; expanded APS-related E2E tests and fixtures.

- **Local/dev environment**
  - Updates to `docker-compose`, local Dockerfiles (Kong, portal, E2E), DB/Keystone init scripts, and `.env.local` / `.nvmrc`.

- **App and dependencies**
  - Small change to the signout page; refreshed dependency lockfiles under `src/`

---
🚀 Feature branch deployment: https://api-services-portal-feature-kc-26.apps.silver.devops.gov.bc.ca


.github/workflows/aps-cypress-e2e.yaml
```diff
@@ -16,16 +16,14 @@ env:
...
-  NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
 jobs:
   cypress-run:
     runs-on: ubuntu-latest
     steps:
       - name: Build GWA API Image
         run: |
-          git clone https://github.com/bcgov/gwa-api.git --branch v1.0.47
+          git clone https://github.com/bcgov/gwa-api.git --branch feature/kc-26
```
danebenal: are we keeping the `feature/kc-26` branch or should this be updated?

Elson9: Once we merge in gwa-api and get a new release tag (likely v1.0.49), this can be updated

danebenal (approved):

rustyjux:
a lot of failed tests (115) in e2e: https://github.com/bcgov/api-services-portal/issues/1370#event-23482753798 - will rerun
When I last left the feature branch there were 5 fails - https://github.com/bcgov/api-services-portal/issues/1310 (3 which must be addressed, not sure why those had not been addressed earlier)

rustyjux:
The test fails were caused by failure on logout, thus a lot of fails since that is after almost every spec.
Root cause would be the change to signout redirect. Should get a better test run now in https://github.com/bcgov/api-services-portal/actions/runs/22973904348

rustyjux:
Alright we have 5 fails in https://github.com/bcgov/api-services-portal/issues/1371, same as I did back in Dec:

Fiddly Cypress issue:

- 07-manage-control/02-rate-limiting.cy.ts - verify rate limit error when the API calls beyond the limit
- 07-manage-control/02-rate-limiting.cy.ts - verify rate limit error when the API calls beyond the limit

These seem like real issues, all 3 show 403s instead of 200s when trying to get activity:

- 15-aps-api/08-namespaces.cy.ts - Get the resource and verify the success code in the response (calling namespaces/{gatewayId}/activity)
- 19-api-v3/03-gateways.ts - POST /gateways (failure when calling `/activity`)
- 19-api-v3/03-gateways.ts - GET /gateways/{gatewayId}/activity (similar to first one here)

Replicating locally, the error for the above 3 is `Missing required scope: <gatewayid>:Namespace.View`. Meanwhile, I can view activity in Portal at `/manager/activity`.

Looking back, I see that in Dec I [added](https://github.com/bcgov/api-services-portal/pull/1369/changes/00ca8bc0969ace87f48d82c75566fd347d38b232) and then [removed](https://github.com/bcgov/api-services-portal/pull/1369/changes/ec6e665c4f22d5aafb92d4b6303c5b18e3486402) `Namespace.View` upon gateway creation, presumably related to this bug. 

My notes captured that users should NOT be able to view `gateway/ns/activity` without Namespace.View explicitly there - but it was possible in KC15 due to bug with resource-based scopes (https://dpdd.atlassian.net/browse/APS-3972).

@ikethecoder do you remember why we removed the change to add that permission on gateway creation? Do we want to A) modify/remove these tests, B) grant the Namespace.View scope manually in the tests so they pass, or C) grant .View on ns creation. ? I'm leaning towards B.

There is also a build error now, which is reproducible locally when running `docker compose --profile testsuite build`, I see @Elson9 is working on it.

ikethecoder:
I can't recall why - but I agree option B.  I think we had talked about a "more correct" solution is to configure the `/manager/activity` endpoint to be accessed with either Namespace.View or Namespace.Manage.

Elson9 (approved):

PR #1386, opened 2026-03-18
APS-4412 - querying currentNamespace deactivates current ns

rustyjux:
Includes:
- fix null current namespace bug
  - Issue: after clicking `Create New Service Account` button, current gateway would become null in dropdown + breadcrumbs
  - Cause: React Query caching and reusing 'currentNamespace' with different fields included for different purposes (not always including `displayName`
  - Fix:  updated service-account-create to use the common `useCurrentNamespace` hook and added scope (needed to SA creation) to that query. Updated gql whitelists + tests

- namespace list is missing display names + publishing enabled/not after going back from a detail page
image

fix: on `detail` page, use query which includes all required attributes

---
🚀 Feature branch deployment: https://api-services-portal-feature-kc-26-fixes.apps.silver.devops.gov.bc.ca

---
🚀 Feature branch deployment: https://api-services-portal-feature-kc-26-fixes.apps.silver.devops.gov.bc.ca

### DIFF
--- a/e2e/cypress/support/auth-commands.ts
+++ b/e2e/cypress/support/auth-commands.ts
@@ -123,9 +123,19 @@ query GetNamespaces {
   const currentNsQuery = `
   query GetCurrentNamespace {
     currentNamespace {
+      id
       name
+      displayName
+      scopes {
+        name
+      }
+      prodEnvId
       org
       orgUnit
+      orgUpdatedAt
+      orgEnabled
+      orgNoticeViewed
+      orgAdmins
     }
   }
 `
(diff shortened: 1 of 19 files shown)


Elson9 (approved):
Looks good to me

rustyjux:
Had many e2e test failures due to invalid query. working to update whitelist (will run in local dev mode to generate required updates)

rustyjux:
A few (14 vs expected 2) failed tests still - https://github.com/bcgov/api-services-portal/issues/1389. I'll try and unpack those but my guess is more invalid queries needing whitelisting

rustyjux:
> A few (14 vs expected 2) failed tests still - #1389. I'll try and unpack those but my guess is more invalid queries needing whitelisting

14 fails was a fluke. Reran and got only two - https://github.com/bcgov/api-services-portal/issues/1390. I think we are good to go.

PR #1391, opened 2026-03-20
APS-4406 - don't get service metrics when value is zero

rustyjux:
The main change here is to avoid errors in batch (which are then passed on to feeder) - the prom queries are changed to only get non-zero values.

Another change is required in `metrics-graph` to avoid shifting the daily metrics to the left if there were zero values at the start of the day.  Rather than deriving the start of the day from the first data point (which especially now, may not be at hour 0), instead we use the day field to calculate the correct start timestamp (midnight of that day).

### DIFF
--- a/feeds/prometheus/index.js
+++ b/feeds/prometheus/index.js
@@ -8,12 +8,12 @@ const log = Logger('prometheus');
 
 const queryRanges = [
   {
-    query: 'sum(increase(kong_http_requests_total[1d])) by (service,code)',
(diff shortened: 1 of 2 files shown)


rustyjux:
Metrics are inconsistent in dev (this change aside) so it is hard to validate the metrics graphs. I reverted the change for metrics-graph and we can add it in future if it proves necessary.

Elson9 (approved):

PR #1417, opened 2026-04-14
APS-4450 add new ca token api for sdx

ikethecoder:
## Overview                                                                                                                             
                                                                                                                                       
  Introduces a new SDX API endpoint for generating one-time-use Step CA tokens used during Certificate Signing Requests (CSR) for runtime groups. Includes the backing service, a local Deno mock for the Step Token API, Docker Compose wiring, and an E2E test.
                                                                                                                                       
  

---
🚀 Feature branch deployment: https://api-services-portal-feature-aps-4450-ca-token.apps.silver.devops.gov.bc.ca

---
🚀 Feature branch deployment: https://api-services-portal-feature-aps-4450-ca-token.apps.silver.devops.gov.bc.ca


ikethecoder:
### Issues & Risks for Review

 | # | Severity | Location | Issue |                                                                                                  
  |---|----------|----------|-------|
  | 1 | Medium | `runtime-group.ts` | `new StepTokenService(process.env.STEP_TOKEN_URL)` — `STEP_TOKEN_URL` may be `undefined` at      runtime if the env var is not set, causing the fetch to throw. Consider asserting or providing a fallback. |                         
  | 2 | Low | `draft-openapi.yaml` vs `main.ts` | OpenAPI spec declares the request field as `name`; the implementation and README both  use `subject`. The spec should be updated to match. |                                                                               
  | 3 | Low | `runtime-group.ts` | `new URL(rg.sdxEndpoint)` is unguarded — a missing or malformed `sdxEndpoint` will throw an unhandled exception. |                                                                                                               
  | 4 | Info | `OrgRuntimeGroupController.ts` | `createContext(request, true)` disables Keystone access control. Appears intentional  for a system operation — worth a brief comment confirming this. |                                                                    
  | 5 | Info | `main.ts` (mock) | JWT secret is hardcoded as `'secret-that-no-one-knows'`. Fine for local dev; the only production  guard is the `STEP_TOKEN_URL` env var pointing away from the mock. |                                                                 
  | 6 | Low | Cypress tests | No sad-path coverage for the new token endpoint. Other runtime group operations have sad-path blocks;  consistency would be good. |

rustyjux (approved):
Simple constrained usage of the API in the new endpoint.

Also I like the single file mock (esp for something simple like this), and appreciate the inclusion of the prompt.

src/controllers/sdx/v1/OrgRuntimeGroupController.ts
```diff
@@ -216,4 +217,33 @@ export class RuntimeGroupController extends Controller {
...
+    @Path() org: string,
+    @Path() name: string,
+    @Request() request: any
+  ): Promise<{ token: string }> {
+    // Create Keystone context with access control disabled
+    const context = this.keystone.createContext(request, true);
+
+    const service = new RuntimeGroupService();
+
+    const token = await service.generateCertSignRequestToken(context, name);
```
Elson9: Should `POST /organizations/{org}/runtime-groups/{name}/tokens` enforce that {name} belongs to (or is explicitly hosted for) {org} before issuing a Step token, instead of resolving the runtime group by name only?

Elson9 (approved):
Approved, with a clarifying question below

src/controllers/sdx/v1/OrgRuntimeGroupController.ts
ikethecoder: Yes :) Thanks.. will update

PR #1423, opened 2026-04-22
APS-4506 introduce connection request

ikethecoder:
## APS-4506: Connection Request Management for SDX                                                                                
                                                                                                                                  
### Overview                                                                                                                      
Introduces a `ConnectionRequest` entity to manage the lifecycle of peer-to-peer connections between SDX consumer and provider subsystems.                                                                                                                       
                                                                                                                                  
### What's New                                                                                                                    

* **`ConnectionRequest` data model** (`src/lists/ConnectionRequest.js`)
A new Keystone list that links a `clientId` (consumer subsystem) to a `serviceId` (provider OAS spec), with `isApproved` and `isActive` flags. On creation, it auto-resolves and stores the client and service organizations via hooks.                        
 
* **Connection API endpoints** (`src/controllers/sdx/v1/OrgConnectionController.ts`)                                                
Two new SDX v1 routes under `/organizations/{org}/connections`:
  - `PUT` — upsert a connection request (requires `System.Manage`)                                                                  
  - `GET` — list all connections for an organization
                                                                                                                                  
* **`ConnectionService`** (`src/services/batch/connection-service.ts`)
Service layer handling upsert, lookup by ID, and list-by-org operations, including org ownership validation and approval guard rails.                     

* **Gateway pattern enforcement** (`sdx-p2p-consumer.ts`, `sdx-p2p-provider.ts`)                                                    
Both the consumer and provider P2P inject flows now validate a `conn_id` input against the `ConnectionRequest` record — checking that `clientId`/`serviceId` match and that the connection is both active and approved before proceeding.                          
                           
* **Authorization** (`src/authz/matrix.csv`)                                                                                        
New rules granting Org Admins read access and Org System Owners create/update/delete access on `ConnectionRequest`.
                                                                                                                                  
* **Supporting additions**
  - `ExtractClientIdFromServiceId` utility in `catalog.ts` to parse service IDs into client IDs                                     
  - `quiet` mode on `EnrichWithRuntimeGroup` to skip namespace errors non-fatally
  - `ConnectionRequest` registered in `data-rules.js` for batch sync with composite key `[clientId, serviceId]`                     
  - E2E tests in `21-sdx-api/v1/06-connection-requests.ts`

---
🚀 Feature branch deployment: https://api-services-portal-feature-aps-4506-access-request.apps.silver.devops.gov.bc.ca

---
🚀 Feature branch deployment: https://api-services-portal-feature-aps-4506-access-request.apps.silver.devops.gov.bc.ca


ikethecoder:
Review

 #: 1                                                                                                                              
  Severity: Bug                             
  File: src/services/batch/connection-service.ts:120                                                                                
  Issue: Wrong field name in listConnectionsByOrganization query
  Detail: Filter uses providerOrganization but the Keystone model defines serviceOrganization. Provider-side queries will silently  
    return zero results.                                                                                                            
  ────────────────────────────────────────
  #: 2                                                                                                                              
  Severity: Bug              
  File: src/services/batch/types.ts                                                                                                 
  Issue: isApproved / isActive typed as string instead of boolean
  Detail: The ConnectionRequest interface declares both as string?. The P2P patterns then do assert.strictEqual(conn.isActive, true,
   
    ...) — 'true' !== true, so approvals will always throw at gateway inject time.
  ────────────────────────────────────────
  #: 3                                                                                                                              
  Severity: Bug
  File: src/services/gateway-patterns/patterns/sdx-p2p-consumer.ts, sdx-p2p-provider.ts                                             
  Issue: conn_id not in requiredParams
  Detail: Both patterns read inputs.conn_id but the requiredParams array is still ['organization', 'client_id', 'service_id']. A
    missing conn_id passes validation and then causes a runtime assertion failure inside getConnectionById.
  ────────────────────────────────────────
  #: 4                                                                                                                              
  Severity: Bug
  File: src/lists/ConnectionRequest.js (hook)                                                                                       
  Issue: Null dereference risk when clientSubsystem not found
  Detail: After the !serviceSpec early-throw, the code accesses clientSubsystem.organization.id before the if (!clientSubsystem)
    guard — if findSubsystemByClientId returns null the hook crashes with a TypeError rather than a clean validation error.
  ────────────────────────────────────────
  #: 5                                                                                                                              
  Severity: Risk
  File: src/services/batch/connection-service.ts:upsertConnection                                                                   
  Issue: syncRecordsThrowErrors called with undefined namespace
  Detail: Third argument (namespace) is undefined. Depending on how feed-worker enforces namespace-scoped access control, this may
    allow cross-org writes.
  ────────────────────────────────────────
  #: 6                                                                                                                              
  Severity: Risk
  File: src/controllers/sdx/v1/OrgServiceController.ts                                                                              
  Issue: Silent empty-body spec upload
  Detail: rawBody fallback is || ''. If the new raw-body middleware skips the request (e.g. readableEnded is true for any proxy
    path), an empty string gets passed to the spec loader with no error raised at the controller boundary.
  ────────────────────────────────────────
  #: 7                                                                                                                              
  Severity: Risk
  File: src/services/batch/connection-service.ts:getConnectionById                                                                  
  Issue: GraphQL variable type mismatch (String vs ID)
  Detail: The query declares $id: String but Keystone ID fields are typed ID. This may silently return empty records rather than a
    hard error, causing the assert.strictEqual(records.length === 1, true, ...) to fire with a misleading message.
  ────────────────────────────────────────
  #: 8                                                                                                                              
  Severity: Risk
  File: src/controllers/sdx/v1/OrgConnectionController.ts                                                                           
  Issue: Authorization asymmetry in upsertConnection
  Detail: When isApproved=true, only the service org is checked. There is no validation that the client org previously submitted the

    request. A service owner can approve a connection for any client without that client having initiated it.
  ────────────────────────────────────────
  #: 9                                                                                                                              
  Severity: Risk
  File: src/lists/ConnectionRequest.js                                                                                              
  Issue: isActive defaults to true on new records
  Detail: New connection requests are created with isActive: true and isApproved: false. Depending on how downstream consumers
    interpret these flags, a not-yet-approved request may still be considered "active" and reach gateway inject flows.
  ────────────────────────────────────────
  #: 10                                                                                                                             
  Severity: Quality
  File: src/controllers/sdx/v1/OrgSubsystemClientController.ts                                                                      
  Issue: Silent subsystem omission
  Detail: EnrichWithRuntimeGroup(..., true) combined with the if (subsystemEntry.runtimeGroup) filter changes the behavior from
    throwing an error to silently dropping subsystems that haven't completed setup. This hides misconfiguration.
  ────────────────────────────────────────
  #: 11                                                                                                                             
  Severity: Quality
  File: src/lists/ConnectionRequest.js (hook)                                                                                       
  Issue: Redundant !serviceSpec null check
  Detail: if (!serviceSpec) throw is checked, then two lines later if (!clientSubsystem || !serviceSpec) is checked again. The
    !serviceSpec branch in the second condition is dead code.
  ────────────────────────────────────────
  #: 12                                                                                                                             
  Severity: Quality
  File: src/services/gwaapi/gwa-service.ts                                                                                          
  Issue: DELETE returns {} — callers need to be verified
  Detail: The fix correctly avoids parsing an empty DELETE response body, but any caller that inspects the return value (previously
    parsed JSON) may now receive an unexpected empty object.
  ────────────────────────────────────────
  #: 13                                                                                                                             
  Severity: Quality
  File: src/api-openapi.js                                                                                                          
  Issue: Raw-body middleware scope
  Detail: The middleware is scoped to /ds/api/sdx/v1 globally, not just the OAS upload route. All SDX v1 endpoints will now buffer
    their full request body into memory twice (raw buffer + body-parser re-parse). Under large payloads or high concurrency this
    doubles memory pressure.

  ---
  Highest priority to fix before merge: items 1, 2, and 3 — they are all hard runtime bugs that will break the connection request
  and gateway inject flows end-to-end.

src/services/batch/connection-service.ts
```diff
@@ -0,0 +1,136 @@
...
+  };
+
+  listConnectionsByOrganization = async (
+    context: Keystone,
+    org: string
+  ): Promise<KeystoneConnectionRequest[]> => {
+    const batchClause = {
+      query: '$org: String',
+      clause:
+        '{ OR: [{ clientOrganization: { name: $org } }, { providerOrganization: { name: $org } }] }',
```
Elson9: Should this be `serviceOrganization` instead of `providerOrganization`?

src/services/batch/types.ts
```diff
@@ -334,6 +334,24 @@ export interface RuntimeGroup {
...
+ *   "isApproved": false,
+ *   "isActive": false
+ * }
+ */  
+export interface ConnectionRequest {
+  slug?: string; // Primary Key
+  clientId?: string;
+  serviceId?: string;
+  isApproved?: string;
+  isActive?: string;
```
Elson9: Should `isApproved` and `isActive` be boolean?

src/services/gateway-patterns/patterns/sdx-p2p-consumer.ts
```diff
@@ -41,6 +42,29 @@ export const SDXP2PConsumerPattern = {
   requiredParams: ['organization', 'client_id', 'service_id'],
```
Elson9: Is `conn_id` missing from requiredParams?

src/services/gateway-patterns/patterns/sdx-p2p-provider.ts
```diff
@@ -49,6 +50,29 @@ export const SDXP2PProviderPattern = {
   requiredParams: ['organization', 'client_id', 'service_id'],
```
Elson9: Is `conn_id` missing from requiredParams?

Elson9 (commented):
Looks good overall, added a couple clarifying questions.

phowells (approved):

src/services/batch/connection-service.ts
ikethecoder: Yes, thx!

Elson9 (approved):

PR #1436, opened 2026-05-05
APS-4533 Delete connection

phowells:
---
🚀 Feature branch deployment: https://api-services-portal-feature-aps-4533-delete-connection.apps.silver.devops.gov.bc.ca

### DIFF
--- a/src/controllers/sdx/v1/OrgConnectionController.ts
+++ b/src/controllers/sdx/v1/OrgConnectionController.ts
@@ -6,7 +6,6 @@ import {
   OperationId,
   Path,
   Put,
-  Query,
   Request,
   Route,
   Security,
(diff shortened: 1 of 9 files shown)


rustyjux (changes requested):
Looks good.
To more thoroughly validate the requirements for this feature, I think we also need an e2e test for the sad path where the connection cannot be deleted because of existing gateway configuration.

Though I see we don't yet have e2e tests which include adding gw config w/ `/pattern`. Am I missing something @ikethecoder?

rustyjux (approved):

src/controllers/sdx/v1/OrgConnectionController.ts
```diff
@@ -83,4 +67,28 @@ export class OrgConnectionController extends Controller {
...
+    const connection = await service.getConnectionById(ctx, id);
+    const organizationMemberId = org.split('-').pop();
+
+    assertEqual(
+      connection.clientId.includes(`.${organizationMemberId}.`) ||
+        connection.serviceId.includes(`.${organizationMemberId}.`),
+      true,
+      'organization',
+      'Not authorized to access this connection request'
+    );
```
ikethecoder: This logic isn't quite right.  The organization member class and member id are in the Organization list - there are some functions to look that up.  Get the subsystem by `clientId` and it will have the organization detail to compare.  And get the OpenAPISpec by 'serviceId' and it will have the organization detail to compare.

ikethecoder (changes requested):

src/controllers/sdx/v1/OrgConnectionController.ts
phowells: Yes, this is not right at all.  I have moved this validation into the ConnectionService, following the pattern of the other endpoints.  I am only testing the org against the clientSubsystem.  I don't think it helps to confirm that the client and spec orgs match when we are deleting.

ikethecoder (approved):

PR #1438, opened 2026-05-07
APS-4452 Org Signing

ikethecoder:
## Summary

  Adds organization-level signing support for the SDX (Secure Data Exchange) platform (APS-4452). Introduces a new endpoint  for orgs to generate signing keys and CSRs against a runtime group's KMS.  Updates the SDX runtime-group pattern to expose KMS routes for the SDX Operator to request keys and CSRs from the Edge Servers.

**BONUS:** Not strictly part of the ticket, but threw this in so we have them:  Extends the P2P consumer/provider gateway patterns with additional upgrades (counter-sign, mTLS, JWT-Keycloak, DPoP, token exchange).

  ## Changes

  - **New `OrgKeys` controller** (`POST /organizations/{org}/keys`) — generates a new key pair and CSR for an organization on
   a runtime group by calling the Edge Server's `/csr` endpoint; returns the result as YAML.
  - **New `sdx-org-keys` workflow** — validates that the runtime group hosts the org, builds the CSR request (country,
  serial, common name, SAN derived from member class/id), and proxies it through the SDX operator edge.
  - **SDX runtime-group pattern** — now provisions a `trust-kms` Gateway Service for `create_key` plus an internal ops route
  (`/edge/{rg}/csr`) on the operator edge so the Portal can drive key creation centrally.
  - **SDX P2P consumer pattern** — adds upgrades for `dpop`, `token` (jwt-keycloak), `counter_sign` (trust-kms request
  signing with `X-Edge-Token`), and `token_exchange`; adds required `source_runtime_group` config (`subsystem` | `pzgw`).
  - **SDX P2P provider pattern** — adds upgrades for `mtls_auth`, `mtls_acl`, `token` (jwt-keycloak), and `counter_sign`
  (trust-kms response signing).
  - **Namespace creation** — reserved domains (e.g. `pzgw.api.gov.bc.ca`) are now included on runtime-group and subsystem
  namespaces, configurable via `SDX_RESERVED_DOMAINS`.
  - **OpenAPI spec loader** — falls back to JSON parsing when YAML parsing fails, with a clearer error if both fail.
  - **v3 API types** — adds `RuntimeGroup` and `ConnectionRequest` models and expands `Subsystem`.
  - **Config** — new env vars: `SDX_PUBLIC_URL`, `SDX_RESERVED_DOMAINS`, `SDX_OPERATOR_EDGE`, `SDX_OPERATOR_EDGE_URL`.

---
🚀 Feature branch deployment: https://api-services-portal-feature-aps-4452-org-signing.apps.silver.devops.gov.bc.ca

---
🚀 Feature branch deployment: https://api-services-portal-feature-aps-4452-org-signing.apps.silver.devops.gov.bc.ca


src/services/gateway-patterns/patterns/sdx-p2p-consumer.ts
```diff
@@ -39,7 +56,13 @@ export interface SDXP2PConsumerPatternData {
  */
 export const SDXP2PConsumerPattern = {
   id: 'sdx-p2p-consumer.r1',
-  requiredParams: ['organization', 'conn_id', 'client_id', 'service_id'],
+  requiredParams: [
+    'organization',
+    'conn_id',
+    'client_id',
+    'service_id',
+    'source_runtime_group',
```
phowells: Do we need the source?  I doesn't look like it is being used

src/controllers/sdx/v1/OrgKeysController.ts
```diff
@@ -0,0 +1,55 @@
...
+    @Request() request: any
+  ): Promise<any> {
+    const context = this.keystone.createContext(request);
+
+    // call the Edge Server endpoint for generating a new key pair
+    const result = await CreateNewKey(context, org, body.runtimeGroupName);
+
+    request.res?.header('Content-Type', 'text/plain; charset=utf-8');
+    request.res?.send(YAML.stringify(result));
+    return '';
```
phowells, line 55: Odd pattern.  Why not set the header and then use the return for the value?

phowells (approved):

docker-compose.yml
```diff
@@ -324,14 +324,28 @@ services:
...
+    container_name: mocks-sdx-edge-api
+    command:
+      - deno
+      - run
+      - --allow-net=:2021,deno.land
+      - --allow-write
+      - --allow-read
+      - /app/main.ts
+    volumes:
+      - ./local/mocks/sdx-edge-api/main.ts:/app/main.ts
```
rustyjux, line 351: seems like you forgot to commit the mock?

src/controllers/sdx/v1/OrgKeysController.ts
ikethecoder: This is a limitation of the version of the TSOA library we are using, and we are not able to update it due to a long story of dependency lock-in :)

rustyjux (approved):

rustyjux (changes requested):
Actually either tests need updating or you need to include the mock. Your new test happy path fails - https://github.com/bcgov/api-services-portal/actions/runs/25848549865

src/services/workflow/sdx-org-keys.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  // call the Edge Server endpoint for generating a new key pair
+  const result = await fetch(routeUrl, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify(body),
+  })
+    .then(checkStatus)
```
Elson9, line 63: If the edge CSR call fails, `checkStatus` throws `IssuerMisconfigError`, which isn’t handled in api-openapi.js - clients get 500 instead of something like 502 with an upstream error message. May be worth adding a handler (or mapping this call to `user-assert` for 422).

Elson9 (commented):
Looks good, only one observation apart from what's already been discussed.

src/services/gateway-patterns/patterns/sdx-p2p-consumer.ts
ikethecoder: Correct - removed.

src/controllers/sdx/v1/OrgKeysController.ts
ikethecoder: Added comment in code to explain

Elson9 (approved):

rustyjux (approved):

PR #1484, opened 2026-05-27
comment instead of creating issue for failed e2e tests on a PR

rustyjux:
I changed a test to quickly demo the test failure behaviour and then restored it.


Elson9 (approved):
Awesome, love it!

PR #1485, opened 2026-05-27
APS-4532 Add activity logging for org admin

rustyjux:
## Updated since review (`2bb99d90`)

Re-review from this commit — everything above it was in the first pass.

- **Org unit activity** — establishment and profile updates; `orgUnit` included in access/profile activity params where relevant
- **Profile snapshots** — full org/org-unit sync field snapshots in activity blobs, not field diffs
- **Key publish activity** — stores GWA deck YAML in the blob on apply; no parsing deck response or key diff
- **Past-tense `Activity.action` verbs** — `registered`, `updated`, `requested`, `published`, `removed` (public catalog filter updated to match)
- **Local dev** — `docker-compose.dev.yml` overlay for WSL2, optional pgadmin (`--profile db-tools`)

---

## Summary

Adds organization-level activity logging for org admin operations and exposes a filtered public SDX catalog feed.

**What gets logged (on success only)**

| Type | Action | Trigger |
|------|--------|---------|
| `Organization` | `registered` | New org via v3 PUT |
| `OrganizationUnit` | `registered` | New unit (v3 PUT or feed sync) |
| `OrganizationProfile` | `updated` | Org or org-unit profile field changes |
| `OrganizationAccess` | `updated` | v2/v3 PUT `…/access` (role grants/revokes per user) |
| `OrganizationCertificate` | `requested` | POST `…/keys` (CSR) |
| `OrganizationKey` | `published` / `removed` | `sdx-keys.r1` pattern apply / remove |

**Public API**

- `GET /ds/api/sdx/v1/catalog/activity` — unauthenticated; filtered by `PUBLIC_ORG_ACTIVITY` (`org-activity-public.ts`)
- Strips `subject_email` from public responses (display name in `params.subject`)
- v2/v3 `GET /organizations/{org}/activity` unchanged (gateway namespace activity, not org admin activity)

**Key publish blob**

On `sdx-keys.r1` apply, the GWA/deck response YAML is stored in the activity blob (e.g. deck `results` showing `creating key …` / `updating key …`). Removes are summarized in `params.detail` (`removed key …`) because there is no deck response

---
🚀 Feature branch deployment: https://api-services-portal-feature-aps-4532-org.apps.silver.devops.gov.bc.ca


src/batch/data-rules.js
```diff
@@ -796,7 +796,15 @@ const metadata = {
...
         type: 'enum',
-        values: ['add', 'update', 'create', 'delete', 'validate', 'publish'],
+        values: [
+          'add',
+          'update',
+          'create',
+          'delete',
+          'validate',
+          'publish',
+          'published',
```
phowells: Is `published` needed here?  It looks like a state, not an action.  I do not see the new org activity code or public activity allow-list using it.

src/batch/feed-worker.ts
```diff
@@ -617,6 +618,15 @@ export const syncRecords = async function (
           childResults,
         };
       } else {
+        if (entity === 'Organization' && localRecord?.name) {
```
phowells: Why would it be missing the name?  Might be better to log it with `unknown` if the name is missing rather than skip it.

src/controllers/sdx/v1/CatalogController.ts
```diff
@@ -45,6 +54,37 @@ export class CatalogController extends Controller {
...
 
+  /**
+   * Retrieve public organization-level activity for the SDX catalog.
+   *
+   * @summary List public organization activity
+   * @param organization Optional organization name filter
+   * @param first Maximum records to return (capped at 100)
+   * @param skip Records to skip for pagination
+   */
+  @Get('/activity')
```
phowells, line 65: One naming concern: `/catalog/activity` reads like a general catalog activity feed, but the implementation and operation id are specifically organization activity (`listPublicOrgActivity`). Should this be scoped in the path as organization activity, e.g. `/catalog/organizations/activity` or `/catalog/organizations/{organization}/activity`?

src/controllers/sdx/v1/CatalogController.ts
```diff
@@ -45,6 +54,37 @@ export class CatalogController extends Controller {
...
+  @OperationId('listPublicOrgActivity')
```
phowells: The existing endpoints that expose organization access/activity details appear to be protected with `@Security(...)` scopes, for example `/organizations/{org}/access` requires `GroupAccess.Manage` and `/organizations/{org}/activity` requires `Namespace.Assign`.

This is public and appears to return the payload without sanitizing actor, username/email, roles, message/context, or blob.

src/controllers/v2/OrganizationController.ts
```diff
@@ -146,7 +144,69 @@ export class OrganizationController extends Controller {
     const groupAccessService = new GroupAccessService(prodEnv.uma2);
     await groupAccessService.login(envConfig.clientId, envConfig.clientSecret);
 
-    await groupAccessService.createOrUpdateGroupAccess(body, ['idir']);
+    const changes = await groupAccessService.createOrUpdateGroupAccess(body, [
+      'idir',
+    ]);
+
+    const ctx = this.keystone.sudo();
+    const orgActivity = new OrgActivityService(ctx, org);
```
phowells: Because this activity service is created with a sudo context, will grant/revoke activity record the authenticated actor, or will `getActor()` fall back to `"system"`?  Should we use a request-backed context for the logging part?

phowells (approved):
Looks great.  Added a few questions/comments.

src/controllers/sdx/v1/CatalogController.ts
ikethecoder: op id is probably more accurately `listPublicActivity`.  It will include other activity - like the changes to subsystems and services.

ikethecoder (commented):
There is a bit of a diff between the org and rbac activity.. one uses `accessAction` to represent the past tense, where as the other hardcodes it in the message.  Haven't looked at the gateway activity, but we should be "correct" and consistent - even if that means potentially changing the gateway activity.

```json
[
  {
    "result": "success",
    "message": "{actor} {accessAction} organization access for {username} ({roles}) on {organization}",
    "params": {
      "action": "revoke",
      "entity": "OrganizationAccess",
      "actor": "system",
      "organization": "ca.bc.gov",
      "username": "aidan.cope@gov.bc.ca",
      "roles": "system-owner",
      "accessAction": "revoked"
    },
    "activityAt": "2026-05-28T16:24:39.422Z"
  },
  {
    "result": "success",
    "message": "{actor} updated organization profile ({changedFields}) for {organization}",
    "params": {
      "action": "update",
      "entity": "OrganizationProfile",
      "actor": "Cope, Aidan CITZ:EX",
      "organization": "ministry-of-food",
      "changedFields": "title"
    },
    "activityAt": "2026-05-28T16:23:54.308Z"
  }
]
```

src/controllers/sdx/v1/CatalogController.ts
ikethecoder: Agreed - for `{actor} {accessAction} organization access for {username} ({roles}) on {organization}`, there may be a `params` that has `username` but we may want to strip it out before being returned.  But a `subjectName` might also be in `pararms` and that is what gets put in the message: `{actor} {accessAction} organization access for {subjectName} ({roles}) on {organization}`.

ikethecoder (commented):
The current RBAC record had me having 'organization-admin' role.  I then updated it to include 'system-owner'.  The activity record showed the below.  This is not quite right, in that I only granted "system-owner" role.  I already had "organization-admin".  Can have a look at the Gateway Administration Access activity (i.e./ `Cope, Aidan CITZ:EX updated Cope, Aidan CITZ:EX permissions: + Content.Publish`).

```json
  {
    "result": "success",
    "message": "{actor} {accessAction} organization access for {username} ({roles}) on {organization}",
    "params": {
      "action": "grant",
      "entity": "OrganizationAccess",
      "actor": "system",
      "organization": "ca.bc.gov",
      "username": "aidan.cope@gov.bc.ca",
      "roles": "organization-admin,system-owner",
      "accessAction": "granted"
    },
    "activityAt": "2026-05-28T16:25:34.083Z"
  }
```

src/batch/data-rules.js
rustyjux: I second guessed this change myself! It is used by gwa-api https://github.com/bcgov/gwa-api/blob/0f9ebd1eb0e020a331e54d416d2c6cd5c2021f00/microservices/gatewayApi/v2/routes/gateway.py#L455. Not having it wasn't blocking the activity from being recorded, but it is more accurate type + docs wise to have it.

src/batch/feed-worker.ts
rustyjux: fair point - name should always be there. simplified.

src/controllers/sdx/v1/CatalogController.ts
rustyjux: So rather than expose email (`username`) we will use Keycloak `display_name` as in namespace access activity?

ikethecoder: You should include email as a param (to be filtered out on the publicActivity), but the `display_name` will be the value in the `message`.

src/controllers/v2/OrganizationController.ts
rustyjux: Good catch, thanks.

src/services/workflow/org-activity.ts
```diff
@@ -0,0 +1,278 @@
...
+
+  async logUpdateOrganizationAccess(
+    success: boolean,
+    data: {
+      username: string;
+      roles: string;
+      accessAction: 'granted' | 'revoked' | 'updated';
+    }
+  ): Promise<void> {
+    const action = data.accessAction === 'revoked' ? 'revoke' : 'grant';
```
Elson9: When `accessAction` is `'updated'` (user in both `changes.granted` and `changes.revoked` on the same PUT), Keystone still stores `action: 'grant'`. The formatted message is fine via `{accessAction}`, and public catalog filtering still matches `grant` + `OrganizationAccess`, but you can’t distinguish updates from net-new grants server-side - namespace access uses `action: 'updated'` directly. Consider mapping `updated` → `action: 'update'` (and adding it to `PUBLIC_ORG_ACTIVITY` if it should be public) for consistency - not blocking if clients rely on `params.accessAction`.

Elson9 (approved):

src/controllers/sdx/v1/OrgGatewaysController.ts
```diff
@@ -168,15 +175,65 @@ export class OrgGatewaysController extends Controller {
...
       artifact
     );
 
+    if (
+      body.pattern === 'sdx-keys.r1' &&
+      !dryRun &&
+      !body.parameters.runtime_group_name &&
+      !body.parameters.client_id
+    ) {
+      const keyDiff =
```
ikethecoder: I'm not sure if field level detail is needed in the activity - or at least, we may want to address in a generic way if we do.

src/controllers/v2/types.ts
```diff
@@ -503,7 +547,7 @@ export interface Activity {
   extRefId?: string; // Primary Key
   type?: string;
   name?: string;
-  action?: "add" | "update" | "create" | "delete" | "validate" | "publish";
+  action?: "add" | "update" | "create" | "delete" | "validate" | "publish" | "published";
```
ikethecoder: was the "published" going to be removed?

src/controllers/v3/OrganizationController.ts
```diff
@@ -103,12 +105,29 @@ export class OrganizationController extends Controller {
...
+
+    const result = await syncRecordsThrowErrors(
+      ctx,
       'Organization',
       body['name'],
       body
     );
+
+    // Profile updates are logged from feed-worker after syncRecords updates.
+    if (!existing && result.result?.startsWith('created')) {
```
ikethecoder, line 123: The "startsWith" seems like it should be more precise rather than interpreting the optional "result".  Is it even needed when the first part confirms it is new?

src/services/workflow/org-activity.ts
```diff
@@ -0,0 +1,358 @@
...
+  ): Promise<void> {
+    return this.recordOrgActivity(
+      success,
+      '{actor} updated organization profile ({changedFields}) for {organization}',
+      {
+        action: 'update',
+        entity: 'OrganizationProfile',
+        actor: this.getActorName(),
+        organization: this.orgName,
+        changedFields: data.changedFields,
```
ikethecoder: I'm not sure if field level detail is needed in the activity - or at least, we may want to address in a generic way if we do.

Collectively, there is quite a lot of code that is determining the changed fields - whether for org, org access or keys.  Wondering whether we should (1) remove field-level differences from the activity in general, or (2) look at a library that provides a "delta" generically, or (3) keep what you have with entity-specific logic for determining what's changed?

ikethecoder (changes requested):

src/controllers/sdx/v1/OrgGatewaysController.ts
rustyjux: what 'field level detail'? if we want to identify add/rotate/delete, we need to parse the keys

src/controllers/v2/types.ts
rustyjux: I don't think so - see https://github.com/bcgov/api-services-portal/pull/1485/changes/BASE..2bb99d90e469c9897a5a6231b1a1f347aaa7bf02#r3319419502 where I replied to Paul on this

src/controllers/v3/OrganizationController.ts
rustyjux: Good catch. I think best is actually
```suggestion
    if (result.result === 'created') {
```
as we see in https://github.com/bcgov/api-services-portal/blob/feature/aps-4329-more-activity/src/controllers/sdx/v1/OrgServiceController.ts#L112
This is what really matters, whereas we could have `!existing` in some funny race conditions or if there is an issue with the getOrganizations lookup

src/controllers/v2/types.ts
ikethecoder: It's not a problem any more :) https://github.com/bcgov/gwa-api/pull/269

Can be removed.

src/controllers/sdx/v1/OrgGatewaysController.ts
ikethecoder: Wondering whether "rotated" was incorrectly included in ticket.. this might be a user experience thing where you remove the last key in your list of keys and add a new key to the top of the list.  But effectively, you have two keys on the go - the "newest" key "0" that new tokens are issued and then the "no longer issued but still available for validation" key "1".  The rotation is just adding a new key to the top, and removing the old key.  But activity would just reflect that a new key was added and an old key was deleted.

Perhaps for Gateway Patterns the result from the gwa apply becomes the detail, as it shows the diff for you (?)

src/services/workflow/org-activity.ts
rustyjux: Re: 2 - Not sure a generic library would get us much. For profiles we probably still need to normalize per-field to
avoid false positives, which is a bunch of the existing code we want to shrink. Or maybe a library handles that too?

Re 1 - this only really applies for profile - we want +/- for group permissions (which syncMembers already produces, added steps for activity are just to pull that out and attached to emails). So yea, we could remove changedFields for profiles.

If we do keep it, I agree we should do a refactor to make a more generic and reusable diffProfile for Subsystem etc. I have a few similar refactors in the next PR to avoid duplication.

I'm happy to explore diff libraries if you think they might actually save a bunch of code or help us to return a fuller before vs after capture (though I don't think we want that).

src/controllers/sdx/v1/OrgGatewaysController.ts
rustyjux: You're right about the UX for key rotation. We can drop 'rotated' and I'll look at using the gwa result as/for the diff/detail.

src/services/workflow/org-activity.ts
ikethecoder: I'm cool with keeping the group permissions diff to keep similar as syncMembers.. and a separate lib might be overkill - but at least having a general function that compares two objects and spits out the diffs should be doable.  I'll leave it up to you on whether you want to leave the profile changedFields in, or take it out and put something better in the next PR.

ikethecoder: Example diff library https://www.npmjs.com/package/diff

rustyjux: I'll take a closer look at this and reusable patterns, with or without a library, in the next PR.
I'll try getting the keys from gwa in this PR.

rustyjux:
@ikethecoder I switched keys to using using the gwa response and believe I've set things up better to handle other gateway pattern activity. I'm going to merge changes into my WIP in APS-4329 and continue work there.

rustyjux:
@ikethecoder really ready for review now

ikethecoder (approved):

PR #1488, opened 2026-05-29
APS-4620 Fix wrong user on namespace access edit when requesterEmail is null

rustyjux:
## Summary

Fixes a bug where editing namespace user access for someone who has not logged into the portal could apply permissions to the wrong Keycloak user.

## Changes

- **Read:** `getPermissionTicketsForResource` falls back to Keycloak Admin API for email when Keystone has no user/email.
- **Write:** Reject empty/whitespace email on grant/update; require exactly one identity-provider match in `lookupUserByEmail`.
- **UI:** Trim email and show an error if grant/update is submitted without one.

## Test plan

- [ ] Grant access to a user who has not logged in; confirm edit modal shows their email after reload.
- [ ] Edit scopes for that user; confirm permissions stay on the correct user.
- [ ] Confirm grant/update with empty email is blocked (UI toast / API error).

---
🚀 Feature branch deployment: https://api-services-portal-feature-aps-4620-email-lookup.apps.silver.devops.gov.bc.ca

### DIFF
--- a/src/nextapp/components/namespace-access/users-access.tsx
+++ b/src/nextapp/components/namespace-access/users-access.tsx
@@ -32,7 +32,7 @@ interface UsersAccessProps {
 }
 
 const EDIT_UNAVAILABLE_MESSAGE =
-  'Email unavailable. Revoke access and re-grant with a valid email.';
+  'Email unavailable. Ask this user to sign in to the API Services Portal with IDIR, then refresh this page. If email is still unavailable, revoke access and re-grant using their email address.';
 
 function canEditUserAccess(item: AccessItem): boolean {
   return Boolean(item.requesterEmail?.trim());
@@ -143,16 +143,6 @@ const UsersAccess: React.FC<UsersAccessProps> = ({
   const handleUpdateAccess = async (form: FormData) => {
     const email = (form.get('email') as string)?.trim();
     const scopes = form.getAll('scopes') as string[];
-
-    if (!email) {
-      toast({
-        status: 'error',
-        title: 'Unable to update user access',
-        description: 'Email is required. Reload the page and try again.',
-        isClosable: true,
-      });
-      return;
-    }
 
     try {
       await update.mutateAsync({
--- a/src/nextapp/components/namespace-access/users-access.tsx
+++ b/src/nextapp/components/namespace-access/users-access.tsx
@@ -31,6 +31,13 @@ interface UsersAccessProps {
   prodEnvId: string;
 }
 
+const EDIT_UNAVAILABLE_MESSAGE =
+  'Email unavailable. Revoke access and re-grant with a valid email.';
+
+function canEditUserAccess(item: AccessItem): boolean {
+  return Boolean(item.requesterEmail?.trim());
+}
+
 const UsersAccess: React.FC<UsersAccessProps> = ({
   resourceId,
   resourceScopes,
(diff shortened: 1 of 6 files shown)


src/services/keycloak/user-service.ts
Elson9, line 73: If a permission references a deleted Keycloak user, or the user has no email, `kcUser.email` will throw and the entire `getPermissionTicketsForResource` query fails. Consider per-requester try/catch or a null check so one stale permission does not break the whole namespace access page.

src/nextapp/components/namespace-access/users-access.tsx
Elson9: If Keycloak also has no email for the user, the edit modal opens with an empty read-only email field and update is blocked. Revoke still works (uses permission IDs). That's safe, but the UX is confusing - consider disabling the Edit button or showing a warning when `requesterEmail` is absent after enrichment.

Elson9 (approved):
Looks great, just a couple minor comments

src/batch/data-rules.js
```diff
@@ -796,7 +796,15 @@ const metadata = {
...
         type: 'enum',
-        values: ['add', 'update', 'create', 'delete', 'validate', 'publish'],
+        values: [
+          'add',
+          'update',
+          'create',
+          'delete',
+          'validate',
+          'publish',
+          'published',
```
ikethecoder: think we said we would remove "published"?

src/services/keycloak/user-service.ts
rustyjux: good call. weird edge case but good to have it covered.

src/nextapp/components/namespace-access/users-access.tsx
rustyjux: updated to disable edit AND show a warning below user name

rustyjux:
Updated to address @Elson9's comments and pass tests (handle multiple matching users / migration). @ikethecoder ready for review

src/services/keycloak/user-service.ts
```diff
@@ -31,6 +31,28 @@ export class KeycloakUserService {
       : undefined;
   }
 
+  /**
+   * Mirrors auth-oauth2-proxy migration logic: the legacy account username is
+   * provider_username@identity_provider, while federated accounts use a GUID.
+   */
+  public isLegacyIdirUser(user: UserRepresentation): boolean {
```
ikethecoder: does this need to be public?

src/services/keycloak/user-service.ts
```diff
@@ -66,19 +88,30 @@ export class KeycloakUserService {
...
+      usersForIdentityProviders.length > 0,
       true,
       `No suitable match for ${identityProviders.join(',')} : ${email}`
     );
-    return user;
+    if (usersForIdentityProviders.length === 1) {
+      return usersForIdentityProviders[0];
+    }
+
+    // During IDIR migration, legacy and federated accounts may share an email.
```
ikethecoder, line 105: Is the scenario where the user has logged in during the legacy period, and then logged in during the current period (i.e./ did the online migration)?  So there are two user records in keycloak for the same email?  Or does the online migration result in one record in keycloak?   Haven't looked at the online migration logic for awhile.

src/nextapp/components/namespace-access/users-access.tsx
```diff
@@ -128,9 +141,19 @@ const UsersAccess: React.FC<UsersAccessProps> = ({
...
   const handleUpdateAccess = async (form: FormData) => {
-    const email = form.get('email') as string;
+    const email = (form.get('email') as string)?.trim();
     const scopes = form.getAll('scopes') as string[];
 
+    if (!email) {
+      toast({
+        status: 'error',
+        title: 'Unable to update user access',
+        description: 'Email is required. Reload the page and try again.',
```
ikethecoder: "Reload the page and try again" seems like a workaround to a bug.. can we update whatever is needed on the page so they don't have to reload the page?

src/nextapp/components/namespace-access/users-access.tsx
```diff
@@ -32,6 +31,13 @@ interface UsersAccessProps {
   prodEnvId: string;
 }
 
+const EDIT_UNAVAILABLE_MESSAGE =
+  'Email unavailable. Revoke access and re-grant with a valid email.';
```
ikethecoder: Do we know how many users are impacted by this?  And how they got given access in the first place without an email?

src/batch/data-rules.js
rustyjux: yea this is gone

src/services/keycloak/user-service.ts
rustyjux: nope — only used internally by lookupUserByEmail. Made it private.

src/nextapp/components/namespace-access/users-access.tsx
rustyjux: Querying the KC admin API, we have 7 federated @idir users and 2 legacy @idir without emails (there are also ~90 BCSC users and 4 @github which don't have email - though these accounts can't do gw admin so maybe not a concern). I didn't look 

Logging into the Portal populates email for federated idir (tested on dev with my own account), though CLI login direct to KC does not.

So we might want to keep logic for handling null email in the UI, but suggest the real fix (the target user should log in to the portal w/ IDIR) in error messages?

rustyjux: Went with my suggestion above.

src/nextapp/components/namespace-access/users-access.tsx
rustyjux: Removed the toast - edit is only reachable when email is set so this was a redundant defensive guard.

ikethecoder (approved):

PR #1492, opened 2026-06-12
APS-4329 - extend SDX/org activity logging

rustyjux:
## Summary

Extends SDX organization activity logging and visibility for APS-4329.

- Logs subsystem lifecycle (created/updated/deleted), service lifecycle (published/removed), organization and subsystem key events, and runtime group lifecycle to org activity and the public catalog feed.
- Adds SDX v1 `GET /organizations/{org}/activity` (System.Manage) to return combined org-scoped SDX activity plus gateway administration events.
- Refactors activity recording around `OrgActivityService` and `OrgActivityResource`; gateway pattern publishes are logged only on confirmed success.
- Splits SDX activity e2e coverage into catalog (`08-catalog-activity.ts`) and org (`09-org-activity.ts`) specs, with shared helpers in `sdx-commands.ts`.

## A couple implementation details
- **refId**:
  - opted to include the Activity.type as part of the refId. This is extra duplicative when you look at Activity.name which includes the type and refId, resulting in `created product[product:my-product-on-gw-7ae20]`
  - this was done for consistency with namespace activities but could be modified
  - refId for keys is messy since multiple keys can be touched per action (e.g. remove). for now refId just takes the first scoped key. 

- **key activity details**:
  - details including the key name(s) are now included in Activity params + message for all actions (not just remove). I realized this was available from the payload and didn't require parsing gwa/deck response.

- **runtime group changes**:
  - For runtime group, only changes to `hostedOrganization`  on an existing runtime group are logged. Updates which change `sdxEndpoint` and `consumerEndpoint` are not

## Test plan

- [x] Unit tests pass (`org-activity`, batch subsystem/oas-service, keystone activity)
- [x] E2e: catalog activity covers subsystem lifecycle, service publish/remove, and key publish/remove on `/catalog/activity`
- [x] E2e: org activity covers runtime group lifecycle and combined org activity on `/organizations/{org}/activity`
- [ ] Verify failed gateway pattern publishes are not recorded as successful activity

---
🚀 Feature branch deployment: https://api-services-portal-feature-aps-4329-clean.apps.silver.devops.gov.bc.ca


docs/postgres/MIGRATION.md
```diff
@@ -18,4 +18,15 @@ alter table public."Organization" add column "publicBodyId" text;
 create unique index organization_publicbodyid_unique
     on public."Organization" ("publicBodyId")
     where "publicBodyId" is not null;
+```
+
+## Activity.filterKey1
```
ikethecoder: At the moment the Activity table has these indexes:
```
SELECT indexname, tablename, indexdef 
FROM pg_indexes 
WHERE schemaname = 'public' and tablename = 'Activity';
      indexname       | tablename |                                  indexdef                                  
----------------------+-----------+----------------------------------------------------------------------------
 Activity_pkey        | Activity  | CREATE UNIQUE INDEX "Activity_pkey" ON public."Activity" USING btree (id)
 activity_actor_index | Activity  | CREATE INDEX activity_actor_index ON public."Activity" USING btree (actor)
 activity_blob_index  | Activity  | CREATE INDEX activity_blob_index ON public."Activity" USING btree (blob)
```

Perhaps just add the indexes on all the filters (as composite keys with the createdAt_utc).. and not get too nuanced with the contents of the filter keys.

src/batch/feed-worker.ts
```diff
@@ -26,6 +26,8 @@ import {
   logOrganizationUnitEstablishedFromRecords,
   logOrganizationUnitProfileChangeFromRecords,
   logOrganizationUnitsFromChildSync,
+  logRuntimeGroupCreatedFromSync,
```
ikethecoder: This component `feed-worker` is meant to be entity-agnostic.  The `data-rules.js` provides the specifics for entities.  I didn't notice from the Org PR that there is entity-specific logic in here.  Might need to see how we can refactor the logging logic so that it does not need to be hardcoded in here.

docs/postgres/MIGRATION.md
rustyjux: Replaced the partial `org:` index with four generic (filterKeyN, createdAt_utc DESC) composite indexes per your suggestion.

src/batch/feed-worker.ts
rustyjux: Yea I noticed that while working on this PR and thought 'not great but Aidan let it go already...' Let's clean er up

I think I will try doing hooks on the keystone lists similar to [Subsystem](https://github.com/bcgov/api-services-portal/blob/feature/aps-4329-clean/src/lists/Subsystem.js#L4).

rustyjux:
@ikethecoder made changes as per your comments. I also tried to cleanup on `org-activity.ts` but it's still a bit of a beast - open to input on how to simplify or split.

src/services/workflow/org-activity.ts
```diff
@@ -477,110 +1116,394 @@ export async function logOrganizationAccessChanges(
...
+  const org = await getOrganizationUnit(context, unitName);
+  if (!org?.name) {
+    logger.error(
+      '[OrgActivity] unit profile change - parent org not found for %s',
+      unitName
+    );
+    return;
+  }
+
+  await new OrgActivityService(context, org.name)
```
Elson9, line 1210: Logs a profile update on every `update`, with no diff check. `logOrganizationActivityFromHook` guards with `hasOrganizationProfileFieldChanges` at lines 1161-1163 - consider the same pattern here to avoid noisy activity from no-op or unrelated batch sync updates.

src/services/workflow/org-activity.ts
```diff
@@ -477,110 +1116,394 @@ export async function logOrganizationAccessChanges(
...
+  if (operation === 'delete') {
+    await orgActivity.logSubsystemDeleted(true, subsystemData);
+    return;
+  }
+  if (operation === 'create') {
+    await orgActivity.logSubsystemCreated(true, subsystemData);
+    return;
+  }
+
+  await orgActivity.logSubsystemProfileChange(true, {
```
Elson9, line 1265: Same concern: every subsystem update logs `logSubsystemProfileChange` even when `name`/`description` are unchanged. A field-change guard (like org profile at 1161-1163) would reduce false-positive activity entries.

src/lists/OrganizationUnit.js
```diff
@@ -37,5 +39,24 @@ module.exports = {
...
-  ]
+  ],
+  hooks: {
+    afterChange: async function ({
+      operation,
+      existingItem,
+      updatedItem,
+      context,
+    }) {
+      if (operation !== 'update') {
```
Elson9, line 50: `afterChange` returns early for non-update operations, so a standalone batch `OrganizationUnit` create never logs establishment. The old `feed-worker` path did this when `!parentRecord?.name`. Units linked via `Organization` sync should still be covered by `logOrganizationUnitsEstablishedFromOrganizationHook`, but please confirm direct top-level unit creates are out of scope - or add a create branch here.

src/services/workflow/org-activity.ts
```diff
@@ -88,16 +400,41 @@ export function buildOrganizationUnitProfileSnapshot(
...
+  );
 }
 
 const SDX_KEYS_PATTERN = 'sdx-keys.r1';
 
 type SdxKeyActivityScope = 'organization' | 'subsystem' | 'runtime-group';
 
+/** Interpret a GWA publish response for activity result logging. */
+export function isGatewayPatternPublishSuccessful(gwaResult: unknown): boolean {
+  if (!gwaResult || typeof gwaResult !== 'object') {
```
Elson9: Ambiguous/missing GWA response (`null`, non-object) is treated as success, so failed publishes could be logged as successful activity.

src/services/workflow/org-activity.ts
```diff
@@ -88,16 +400,41 @@ export function buildOrganizationUnitProfileSnapshot(
...
+
+  if (typeof result.failed === 'number') {
+    return result.failed === 0;
+  }
+
+  if (Array.isArray(result.results)) {
+    return result.results.every((item) => item.status !== 'failed');
+  }
+
+  return true;
```
Elson9: Fallback `return true` when the response has no `failed` count or `results` array - same risk. Consider defaulting to `false` (or logging a warning) when the response shape is unrecognized.

src/services/workflow/org-activity.ts
```diff
@@ -477,110 +1116,394 @@ export async function logOrganizationAccessChanges(
...
+export async function getCombinedOrganizationActivity(
+  context: any,
+  org: string,
+  first: number = 20,
+  skip: number = 0
+): Promise<Activity[]> {
+  const cappedFirst = first > 100 ? 100 : first;
+  const fetchLimit = Math.min(cappedFirst + skip, 100);
+
+  const orgRecords = await getOrgActivity(
```
Elson9, line 1488: Calls `getOrgActivity(..., publicOnly: false)`, so failed gateway publishes appear in the admin feed. Public catalog excludes them via `result_not: 'failed'` in `activity.ts` ~252. Intentional for audit, or should this endpoint also filter failed results?

Elson9 (commented):
Great work overall. Just a few minor comments.

src/services/workflow/org-activity.ts
rustyjux: org unit updates run through batch which only updates in the case of changes - https://github.com/bcgov/api-services-portal/blob/feature/aps-4329-clean/src/batch/feed-worker.ts#L525-L526
so I'd say not needed here.

src/services/workflow/org-activity.ts
rustyjux: I’m less worried here: subsystems aren’t batch-fed like orgs, and SDX upserts go through syncRecords, which already returns no-change when nothing’s different. The only real profile update path is description (other fields can't change). So I don’t think we’d get noisy activity.

src/lists/OrganizationUnit.js
rustyjux: All real creates go through org sync and are already covered. Direct org unit creation is via PUT /feed/OrganizationUnit is dev/manual only and unused by feeders

src/services/workflow/org-activity.ts
rustyjux: Intentional for audit, and intentionally not public.

src/services/workflow/org-activity.ts
rustyjux: Good catch — updated `isGatewayPatternPublishSuccessful` to take the gateway action (`apply` vs `remove`).

For apply, null/empty shapes --> don't log as success.

For remove, DELETE doesn't return a deck body (`GWAService` stubs `{}`), so no failure signals in the response --> success.

HTTP errors already throw via `checkStatus` before we get here — this helper is really for the partial-failure-in-200-body case.

src/services/workflow/org-activity.ts
rustyjux: Valid. Addressed by the previous comment.

ikethecoder (approved):

PR #1494, opened 2026-06-29
APS-4038 Validate uploaded OpenAPI specs with configured OAS ruleset

phowells:
---
🚀 Feature branch deployment: https://api-services-portal-feature-aps-4038.apps.silver.devops.gov.bc.ca


src/services/workflow/openapi-spec-validation-service.ts
```diff
@@ -0,0 +1,260 @@
...
+
+    const res = await fetch(url, {
+      method: 'POST',
+      body: spec,
+      headers: {
+        'Content-Type': contentTypeForSpec(spec),
+      },
+    });
+
+    if (!res.ok) {
```
Elson9, line 112: When the validation API is down or returns a non-2xx, this throws a generic `Error`, which bubbles up as a 500 from the global handler. That’s a bit misleading for the client - the spec might be fine, but the dependency failed. Consider a dedicated error type (e.g. `ServiceUnavailableError`) and return 503 with a clear message like “OAS validation service unavailable.” Same applies to the `GET /versions` failure path at lines 177-187.

src/services/workflow/openapi-spec-validation-service.ts
```diff
@@ -0,0 +1,260 @@
...
+      version
+    )}/rulesets/${encodeURIComponent(ruleset)}/validations`;
+
+    logger.debug(
+      'Validating OpenAPI spec using version %s and ruleset %s',
+      version,
+      ruleset
+    );
+
+    const res = await fetch(url, {
```
Elson9: There’s no timeout on these fetch calls (also on `GET /versions` at line 170). If the validation API hangs, OAS uploads will hang too. Worth adding an `AbortSignal` with a reasonable timeout (e.g. 30s) and treating timeout as a 503/dependency failure rather than leaving the request open indefinitely.

Elson9 (approved):
Looks good, just a couple minor comments

rustyjux (changes requested):
Implementation looks fine but there a few problems with testing:

1. e2e tests call on the external dev Validation API. Better to include in docker compose setup to allow for more reliable local + CI runs, along the lines of what we do with `gwa-api`. Though it would be nice to not need to manually clone and build, rather the docker compose service should default to a published image on GHCR, with potential to override for local build. Then point the portal at http://oas-validation-api.localtest.me:8080.
2. SDX e2e tests `/cypress/tests/21-*/` were not running - I fixed this in https://github.com/bcgov/api-services-portal/commit/d02772d5588247c7d5bf48216b9c21ff066fc343. Running the tests in `02-oas-services` locally which will hit the new validation path, I saw 500 errors returned but didn't delve into what was erroring on the dev instance
3. e2e workflow is failing at `docker compose --profile testsuite build` (before tests run). `local/portal/Dockerfile.E2E` runs `nyc instrument --in-place` on the TypeScript sources before `npm run build`, leading to the error. A minor refactor should address this
```
90.94 services/workflow/openapi-spec-validation-service.ts(1844,3): error TS2376: A 'super' call must be the first statement in the constructor when a class contains initialized properties, parameter properties, or private identifiers.
91.01 ERROR: "ts-build" exited with 2.
```
4. build (outside of e2e) is failing due to missing env var in https://github.com/bcgov/api-services-portal/blob/feature/aps-4038/.github/workflows/ci-build-deploy.yaml#L373. To make this a little more confusing, you had a run fail yesterday to due a flaky network issue w/ openshift (which we can ignore here / rerun the action when needed), but I recorded a failure here: https://github.com/bcgov/api-services-portal/actions/runs/28456592924/job/84332967975

src/services/workflow/openapi-spec-validation-service.ts
phowells: Updated

src/services/workflow/openapi-spec-validation-service.ts
phowells: Updated

rustyjux (approved):
Looks gouda. e2e should pass (or rather only the 2 rate limit tests should fail). Wait for the result before merging

ikethecoder (approved):

PR #1505, opened 2026-07-29
APS-4652 Fix SDX peer-to-peer gateway resolution

phowells:
## Summary

- Resolve `sdx-p2p-consumer.r1` authorization through the subsystem gateway using `parameters.clientId`.
- Resolve `sdx-p2p-provider.r1` authorization through the service gateway using `parameters.serviceId`.
- Add both peer-to-peer patterns to the valid-pattern list.
- Leave gateway resolution unresolved when the required identifier is missing.

## Testing

- Verified consumer and provider gateway resolution.
- Verified missing identifiers do not select another gateway.
- Verified existing subsystem, runtime-group, service, and connection mappings.
- `git diff --check` passes.

---
🚀 Feature branch deployment: https://api-services-portal-feature-aps-4652-p2p-gateway.apps.silver.devops.gov.bc.ca

### DIFF
--- a/src/auth/auth-sdx-middle.ts
+++ b/src/auth/auth-sdx-middle.ts
@@ -54,6 +54,10 @@ export class AuthMiddle {
       return await this.lookupSubsystemGateway(org, body.parameters.clientId);
     }
 
+    if (pattern === 'sdx-p2p-consumer.r1' && body.parameters?.clientId) {
+      return await this.lookupSubsystemGateway(org, body.parameters.clientId);
+    }
+
     if (
       pattern === 'sdx-runtime-group.r1' &&
       body.parameters?.runtimeGroupName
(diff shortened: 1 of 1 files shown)


rustyjux:
I think you should target this against `dev` - James' WIP is not related and better keep these changes isolated

src/auth/auth-sdx-middle.ts
```diff
@@ -54,6 +54,10 @@ export class AuthMiddle {
       return await this.lookupSubsystemGateway(org, body.parameters.clientId);
     }
 
+    if (pattern === 'sdx-p2p-consumer.r1' && body.parameters?.clientId) {
```
rustyjux, line 57: Could simplify and use `||` (OR) here and for service/p2p-provider
```
if (
  (pattern === 'sdx-subsystem.r1' || pattern === 'sdx-p2p-consumer.r1') &&
  body.parameters?.clientId
) {
  return await this.lookupSubsystemGateway(org, body.parameters.clientId);
}
```

rustyjux (approved):
See comment + suggestion to change target to `dev`

phowells:
Aidan has confirmed that this is not the correct solution so I will cancel this PR.

PR #1506, opened 2026-07-31
Feature/e2e sdx modular harness

ikethecoder:
---
🚀 Feature branch deployment: https://api-services-portal-feature-e2e-sdx-modular-harness.apps.silver.devops.gov.bc.ca

PR #1507, opened 2026-07-31
PLAT-001: reject unsupported organization-access roles

ikethecoder:
## Summary

`createOrUpdateGroupAccess` filtered its sync loop down to `OrganizationRoles` (`organization-admin`, `system-admin`) but never validated the caller's submitted role names against that set first. Submitting an unsupported role (e.g. `system-owner` — a real `PredefinedRolePermissions` key, just not a supported *organization*-level role) silently matched nothing in the loop; the call still returned `204` and granted nothing, with no way for the caller to tell the write was a no-op.

Now validates every submitted role up front and throws naming the unsupported value(s).

## Test plan

- [x] `node e2e-sdx/run.js --test plat-001` (from the paired harness PR #1506): before the fix, `put-organization-access` with `roles: ["system-owner"]` returned `204` and logged `CONFIRMED`. After rebuilding with this fix and re-running, the same call now returns a 4xx naming the role and the scenario logs `RESOLVED`.

### DIFF
--- a/src/services/org-groups/group-access.ts
+++ b/src/services/org-groups/group-access.ts
@@ -46,12 +46,12 @@ export class GroupAccessService {
     // ["system-owner"]` (a valid *system-level* role, not a supported
     // *organization*-level one) 204'd without assigning anything. Fail
     // fast and name the offending role(s) instead.
+    const submittedRoles: string[] = (groupMembership.members ?? []).reduce(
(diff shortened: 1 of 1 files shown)


phowells (approved):
The role validation appears correct and is performed before changes are applied.

Looks good.

ikethecoder:
Closing: superseded / no longer needed.

PR #1509, opened 2026-07-31
ERR-014: make subsystem gateway registration an idempotent reconcile

ikethecoder:
## Summary

`checkNamespaceAvailable`'s unconditional throw meant repeating an already-successful registration (or retrying after a partial failure) always 422'd with "Namespace already exists", with no supported way to update `perm-domains`/`perm-route-paths`/etc. on an existing namespace.

`CreateNamespace` now checks existence via a new non-throwing `NamespaceService.namespaceExists` instead: on an existing namespace it looks up the matching UMA resource set by name (`findResourceByName`) rather than recreating it and its permission tickets, then always reconciles the Keycloak group's attributes (previously gated behind `created`, so even a fixed create-only bug would have left attribute updates silently no-op'ing on an existing group).

Live re-verification of the first commit surfaced a second, downstream create-only step: `createSDXNamespace` unconditionally called `createUmaPolicy` for the `sdx-provisioner` client after `CreateNamespace` returns — which 409-conflicted on retry. The obvious fix (checking existence via `listPolicies` first) itself hit a separate, pre-existing gap: that endpoint rejects the same bearer token the create/update POST endpoints accept, in this environment. The final commit instead tries create first and treats a 409 as already-correctly-configured, avoiding the broken endpoint.

3 commits, in order: the core reconcile, the UMA-policy follow-on, and the endpoint-workaround correction.

## Test plan

- [x] `node e2e-sdx/run.js --test err-014` (from the paired harness PR #1506): before the fix, repeating `register-subsystem-gateway` with identical parameters returned `422 Namespace already exists`. After rebuilding with all 3 commits and re-running, the same repeat call now returns `200`/reconciled.
- [x] `node e2e-sdx/run.js` (happy-path) confirmed no regression after each commit.

### DIFF
--- a/src/services/org-groups/namespace.ts
+++ b/src/services/org-groups/namespace.ts
@@ -150,6 +150,11 @@ export class NamespaceService {
     assert.strictEqual(groupExists, false, 'Namespace already exists');
   }
 
+  /** Non-throwing existence check - see CreateNamespace's reconcile path (ERR-014). */
+  async namespaceExists(ns: string): Promise<boolean> {
+    return await this.groupService.hasGroup('ns', ns);
+  }
+
   async listAssignedNamespacesByOrg(org: string): Promise<OrgNamespace[]> {
     const groups = await this.groupService.getGroups('ns', false);
     assert.strictEqual(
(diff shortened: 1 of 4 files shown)


phowells (changes requested):
The reconcile path appears to identify an existing registration only by namespace name. Once that group is found, we overwrite its org, perm-runtime-group, perm-data-plane, domains, and route paths without first confirming that it represents the same subsystem gateway registration.

Could we validate the immutable ownership/association fields before reconciling? An identical retry should succeed, but an existing namespace associated with a different organization, runtime group, or data plane should still be rejected rather than reassigned.

PR #1510, opened 2026-07-31
ERR-015: distinguish validator engine failures from real unavailability

ikethecoder:
## Summary

`openapi-spec-validation-service.ts`'s `createValidation` (and the sibling `getSupportedVersions`/`getSupportedRulesets`) mapped *every* non-2xx response from the OAS validator to `OpenAPISpecValidationServiceUnavailableError`, erasing the difference between an unreachable validator and a reachable one that rejected the request (a Spectral engine crash, a malformed-body 400, ...). Confirmed live: a spec triggering the validator's Spectral null-guard crash returns 500 and was reported to callers as "OAS validation service unavailable" — directing operators toward availability/network diagnostics for an engine-side problem.

Non-2xx validator responses now throw a new `OpenAPISpecValidationEngineError`, preserving the real status/statusText/body; `OpenAPISpecValidationServiceUnavailableError` is reserved for actual connectivity failures (timeouts/aborts, an empty supported-versions list). `api-openapi.js`'s error-mapping middleware reports the new error as `502 validation_engine_error` with the preserved diagnostic fields, instead of a generic `503`. Updates the two existing unit tests that asserted the old masked behavior.

## Test plan

- [x] `node e2e-sdx/run.js --test err-015` (from the paired harness PR #1506): before the fix, a spec crashing the validator's Spectral engine returned `503 validation_service_unavailable`. After rebuilding with this fix, the same spec now returns `502 validation_engine_error` with the real "Spectral validation engine internal error" body.
- [x] Updated `src/test/services/workflow/openapi-spec-validation-service.test.js`'s two tests that asserted the old masked behavior.

### DIFF
--- a/src/api-openapi.js
+++ b/src/api-openapi.js
@@ -212,6 +212,21 @@ class ApiOpenapiApp {
           code: 'validation_service_unavailable',
           message: 'OAS validation service unavailable',
         });
+      } else if (err?.name === 'OpenAPISpecValidationEngineError') {
+        // ERR-015: the validator was reachable but rejected the request
+        // itself (e.g. an internal engine crash, a malformed-body 400) -
+        // distinct from genuine unavailability above. Preserve its real
+        // status/diagnostic text instead of reporting a generic 503.
+        logger.error(
+          `Caught OpenAPI Spec Validation Engine Error for ${req.path}:`,
+          err.message
+        );
+        return res.status(502).json({
+          code: 'validation_engine_error',
+          message: err.message,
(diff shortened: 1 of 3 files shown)


phowells (changes requested):
We are returning the validator’s raw response body to the client through both message and validatorBody. Because this is arbitrary upstream error text, it could expose stack traces or other internal implementation details.

Could we retain the full body in the server logs while returning a stable, sanitized error response to the client? We could still include the upstream status and a request/correlation identifier for troubleshooting.

PR #1511, opened 2026-07-31
ERR-019: apply the SDX publication transform to OAD retrieval endpoints

ikethecoder:
## Summary

Both the public catalog OAD endpoint (`CatalogController.getOASServiceSpec`) and the organization-scoped one (`GatewayServiceController.getOrganizationServiceSpec`) returned exactly the stored spec — validation-enriched (`x-csbc-api-standard`/`-ruleset`, added at upload time by `openapi-spec-loader.ts`) but otherwise untouched. The SDX API Standard describes a further publication transform: environment-specific `servers`, no provider `info.contact`, and an SDX `externalDocs` entry — none of which were applied, so the live Widget OAD still exposed its original `/api/v1` server and provider contact info with no SDX `externalDocs`.

Adds `openapi-spec-publisher.ts`'s `publishOpenApiSpec`, applied at *read* time (the stored copy is untouched, keeping "what was validated" and "what's published" distinguishable): strips `info.contact`, rewrites `servers` to the SDX-published endpoint using the same `/sdx/0/{clientId}` route convention already established in `services/utils.ts`, and adds `externalDocs` pointing at SDX documentation.

OAuth endpoint rewriting (also described by the Standard) is intentionally left as a documented gap — there's no established per-service convention for it in this codebase to build on, unlike the route prefix.

## Test plan

- [x] `node e2e-sdx/run.js --test err-019` (from the paired harness PR #1506): before the fix, a service registered with a distinctive `servers` URL and `info.contact` retained both unchanged on retrieval, with no `externalDocs`. After rebuilding with this fix, the retrieved spec no longer exposes the original `servers` URL or `info.contact`, and now carries an `externalDocs` entry alongside the pre-existing `x-csbc-api-standard` enrichment.

### DIFF
--- a/src/controllers/sdx/v1/CatalogController.ts
+++ b/src/controllers/sdx/v1/CatalogController.ts
@@ -43,6 +43,7 @@ import {
 import { KeystoneService } from '../../ioc/keystoneInjector';
 import assert from 'assert';
 import { ResourceScope } from '../../../services/workflow/openapi-spec-loader';
+import { publishOpenApiSpec } from '../../../services/workflow/openapi-spec-publisher';
 
 interface MissingCredentialsJSON {
   code: 'credentials_required' | 'invalid_token';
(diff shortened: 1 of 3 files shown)


phowells (approved):
Looks good.

PR #1512, opened 2026-07-31
ERR-020: require policyVersion when creating a connection request

ikethecoder:
## Summary

`src/lists/ConnectionRequest.js` marks `policyVersion` required at the Keystone list layer, but the public controller/DTO layer (`ConnectionRequestInput.policyVersion`) treats it as optional — so omitting it wasn't caught until deep in the batch-create path, which only reported a generic `create-failed`/"invalid mutation" with no field-specific explanation.

`upsertConnection` is shared by create and every patch-style follow-up call (consumer-open, provider-open, approve, deactivate), so `policyVersion` can't be required unconditionally — only when no connection exists yet for the `clientId`/`serviceId` pair. `ConnectionService.upsertConnection` now checks for an existing connection first; if none exists (a create), `policyVersion` is required via a field-specific `ValidateError`.

## Test plan

- [x] `node e2e-sdx/run.js --test err-020` (from the paired harness PR #1506): before the fix, creating a connection without `policyVersion` returned a generic `400 create-failed`/"invalid mutation" with no field name. After rebuilding with this fix, the same request now returns `422` with `fields.policyVersion.message` naming the missing field.
- [x] `node e2e-sdx/run.js` (happy-path) confirmed no regression — all the update-style connection calls (which never send `policyVersion`) are unaffected.

### DIFF
--- a/src/services/batch/connection-service.ts
+++ b/src/services/batch/connection-service.ts
@@ -18,7 +18,7 @@ import { ConnectionRequestInput } from '../../controllers/sdx/v1/types';
 import { SubsystemService } from './subsystem';
 import { OpenAPISpecService } from './oas-service';
 import { strict as assert } from 'assert';
-import { assertEqual } from '../../controllers/ioc/assert';
+import { assertEqual, assertIsDefined } from '../../controllers/ioc/assert';
 import { ProvisionerService } from '../provisioner';
 
 const logger = Logger('batch.connection');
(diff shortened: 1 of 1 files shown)


phowells (approved):
assertIsDefined only rejects undefined, so policyVersion: null or policyVersion: "" can bypass this validation and fail later with a less useful error.

Could we validate that policyVersion is a non-empty string when creating the connection request, rather than only checking that the property is present?

phowells (changes requested):
assertIsDefined only rejects undefined, so policyVersion: null or policyVersion: "" can bypass this validation and fail later with a less useful error.

Could we validate that policyVersion is a non-empty string when creating the connection request, rather than only checking that the property is present?

PR #1513, opened 2026-07-31
ERR-023: populate the R0 requester even when requesterDetails is omitted

ikethecoder:
## Summary

`upsertConnection` only overwrote `requesterDetails.requester` with the authenticated caller's identity when the incoming mutation already contained a truthy `requesterDetails` object. Omitting it entirely (the documented DevHub example) instead let `ConnectionRequest.js`'s invalid persistence default (`'{}'`) through untouched — the `SDX.R0.00` Cedar context schema requires a nested `requester.name` whenever `requesterDetails` is present at all, so that default fails policy evaluation at activation with `"expected the record to have an attribute 'requester', but it does not"`, confirmed live in the provisioner logs.

Now populated unconditionally whenever `policyVersion` is `SDX.R0.00`, regardless of whether the caller sent `requesterDetails`.

## Test plan

- [x] `node e2e-sdx/run.js --test err-023` (from the paired harness PR #1506): before the fix, a connection created without `requesterDetails` activated with the provisioner logging a Cedar policy check failure ("expected the record to have an attribute `requester`, but it does not"). After rebuilding with this fix, the same connection now activates with the provisioner logging `"Policy check passed"`.

### DIFF
--- a/src/controllers/sdx/v1/OrgConnectionController.ts
+++ b/src/controllers/sdx/v1/OrgConnectionController.ts
@@ -58,11 +58,22 @@ export class OrgConnectionController extends Controller {
   ): Promise<BatchResult> {
     const ctx = this.keystone.createContext(request);
 
-    // For R0 policy, force the requester details to be the user making this request
-    if (input.policyVersion === 'SDX.R0.00' && input.requesterDetails) {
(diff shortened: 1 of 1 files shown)


phowells (approved):
Looks good

PR #1514, opened 2026-07-31
ERR-024: await and propagate the provisioner call in postConnectionRequestChangeEvent

ikethecoder:
## Summary

`postConnectionRequestChangeEvent` was declared `async` returning `Promise<void>`, but its `fetch(...).then(...).then(...).catch(...)` chain was never returned or awaited — the method resolved as soon as the request was *started*, not when it completed, and any failure was only ever logged. Its sole caller, `ConnectionRequest.js`'s `afterChange` hook, already does `await provisionerService.postConnectionRequestChangeEvent(...)` — that await was meaningless against a promise that resolved early.

Confirmed live: setting `isActive: true` on a connection whose downstream policy evaluation then failed still returned HTTP 200/"updated" immediately, with the failure only visible ~2s later in the provider organization's activity feed.

Now returns/re-throws the fetch chain, so the existing `await` in the hook genuinely waits for the provisioner call and a failure propagates to the mutation response instead of being swallowed. Does **not** additionally populate `provisionerStatus` with pending/success/failed states (the errata's other suggested option) — doing that safely from this list's own `afterChange` hook needs a way to write back to the record without re-triggering the same hook, which this change doesn't add.

## Test plan

- [x] `node e2e-sdx/run.js --test err-024` (from the paired harness PR #1506): activates a connection submitted with `policyVersion: SDX.R1.00` without R1's required requester shape, so provisioning is guaranteed to fail independent of any other fix. Before this fix, the activation call returned `200 updated` regardless. After rebuilding with this fix, the same activation now returns `400 update-failed` with the provisioner's `500 Internal server error` reason.

### DIFF
--- a/src/services/provisioner/provisioner-service.ts
+++ b/src/services/provisioner/provisioner-service.ts
@@ -92,7 +92,7 @@ export class ProvisionerService {
   public async postConnectionRequestChangeEvent(
     connection: ConnectionRequest,
     action: 'apply' | 'delete'
-  ): Promise<void> {
+  ): Promise<ConnectionRequestChangeEventResponse> {
     // credentials will be using the gwa admin
 
     const payload = {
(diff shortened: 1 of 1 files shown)


phowells (changes requested):
Because this call is made from afterChange, the connection update has already been persisted before the provisioner is invoked. Rethrowing the provisioner error makes the mutation return update-failed, but it does not roll back isActive.

This appears to leave the connection marked active even though provisioning failed. Could we either move the provisioner operation before the state change, explicitly revert the state when provisioning fails, or record a provisioning status so the persisted state accurately represents the outcome?

PR #1516, opened 2026-07-31
ERR-029: enforce per-operation OAS scopes instead of one blanket service plugin

ikethecoder:
## Summary

`sdx-service.r1` never read `op.scopes` when generating each operation's Kong route; the only runtime auth plugin it could attach was one `jwt-keycloak` on the whole service (`upgradeToJWTKeycloak`, gated on the `token` upgrade), using a single statically-configured `token.scope` with no per-operation differentiation. `ServiceCatalogOperation.scopes` existed purely as OAS metadata with no enforcement behind it — any caller holding a valid token could reach every operation regardless of its declared scope.

`jwt-keycloak` now attaches per generated route instead of once at the service level, with `scope` computed from that operation's own OAS-declared scopes (falling back to the static `token.scope` config for operations that declare none). The synthetic `/hello` health-check route (which short-circuits via `request-termination` before reaching any operation) is not included, since it has no OAS operation/scopes to enforce.

## Test plan

- [x] `node e2e-sdx/run.js --test err-029` (from the paired harness PR #1506): registers a service whose OAS declares distinct `widget:read`/`widget:write` scopes on two operations, then previews `sdx-service.r1` with the `token` upgrade enabled. Before the fix, the preview attached one blanket `jwt-keycloak` plugin with no scope differentiation. After rebuilding with this fix, the preview now generates two separate `jwt-keycloak` plugins with `scope: "widget:read"` and `scope: "widget:write"` respectively.

### DIFF
--- a/provisioner-api/src/services/gateway-patterns/sdx-service.ts
+++ b/provisioner-api/src/services/gateway-patterns/sdx-service.ts
@@ -145,6 +145,22 @@ export class SDXServicePattern implements PatternProcessor {
           },
           protocols: inputs.useSni === 'false' ? ['http'] : ['https'],
           strip_path: false,
+          // ERR-029: OAS-declared per-operation scopes were metadata only -
+          // the only runtime auth plugin generated was one blanket
+          // jwt-keycloak on the whole *service*, using a single static
+          // `token.scope` upgrade config with no per-operation
+          // differentiation. This attaches jwt-keycloak per *route*
+          // instead, once the `token` upgrade is enabled, restricting each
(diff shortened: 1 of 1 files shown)


phowells (commented):
Moving jwt-keycloak from the service to the generated operation routes also removes authentication from the synthetic /hello route. Previously, the service-level plugin protected that route whenever the token upgrade was enabled.

Is /hello intentionally meant to become publicly accessible? If not, it should receive a route-level jwt-keycloak plugin as well, probably using the configured fallback token.scope.

PR #1522, opened 2026-08-05
SP136: route runtime group cert-signing tokens through the provisioner

ikethecoder:
## Summary

Runtime group certificate-signing token generation called step-ca directly
from the portal via a single global `STEP_TOKEN_URL`, ignoring the runtime
group's environment. Every environment's runtime groups got their tokens
from the same step-ca instance.

- Add a per-environment `ca_token_url` to the provisioner's environment
  config, and a new provisioner operation
  (`POST .../runtime-groups/{name}/environments/{env}/cert-token`) that
  resolves and calls the right step-ca for that environment.
- The portal now calls the provisioner instead of step-ca directly
  (`RuntimeGroupService.generateCertSignRequestToken`); removed the now-dead
  direct step-ca client and `STEP_TOKEN_URL`.
- Local dev stack: added a second CA mock and a `cyp` environment so
  per-environment routing is actually observable locally, each mock on its
  own internal port (2020 / 2022) rather than sharing one and being
  distinguished only by host port mapping.
- New `e2e/cypress/tests/99-sp136` suite asserts each runtime group's token
  request lands on its own environment's CA (verified failing on the old
  code, passing after the fix).

## Test plan

- [x] New `e2e/cypress/tests/99-sp136` suite passes
- [x] Existing `e2e/cypress/tests/21-sdx-api/v1/04-runtime-groups.ts` suite
      passes (no regression)
- [x] Verified the new test fails against the pre-fix code and passes
      against the fix


rustyjux (approved):
looks good. cypress test directory 99 is a little weird not sure if you will address that in the merge into dev

provisioner-api/src/controllers/runtime-groups-controller.ts
```diff
@@ -24,4 +35,17 @@ export class RuntimeGroupsController {
...
   }
+
+  async generateCertToken(
+    input: GenerateCertTokenInput
+  ): Promise<CaTokenResponse> {
+    this.logger?.debug(
+      { org: input.org, name: input.name, environment: input.environment },
+      'RuntimeGroupsController.generateCertToken'
+    );
+    return await this.services.caToken.generateCertToken(
```
phowells, line 42: Could you please clarify where the `org` and `name` parameters are validated before this token is minted? `createCsr` delegates to `SdxOperatorService`, which looks up the runtime group and derives its certificate fields, while this path does not use `org` or `name` and forwards the caller-provided `subject` and `san` directly. If the provisioner endpoint can be called independently, it appears that a caller could request a CA token for names unrelated to the runtime group in the URL. Would it make sense to validate or derive these values here as well?

provisioner-api/src/clients/ca-token/client.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+}
+
+async function parseJson<T>(res: Response): Promise<T> {
+  const text = await safeText(res);
+  if (text === '') return undefined as T;
+  try {
+    return JSON.parse(text) as T;
+  } catch {
+    return text as T;
```
phowells, line 146: Could the successful response be validated before it is returned as `CaTokenResponse`? `parseJson` currently returns `undefined` for an empty body and casts non-JSON text to `T`, so a 2xx CA response without `{ token: string }` can escape this client and fail later during response serialization or destructuring rather than being reported as a CA `BadGatewayError`. Would an explicit shape check here make that failure clearer?

provisioner-api/src/routes/runtime-groups.ts
```diff
@@ -50,4 +55,43 @@ export const registerRuntimeGroupsRoutes: FastifyPluginAsyncTypebox = async (
         request: req.body,
       })
   );
+
+  app.post(
```
Elson9, line 58: the new certificate-token route declares security only in OpenAPI. The Fastify app registers no authentication middleware, so any caller that can reach the provisioner can POST to this endpoint without a bearer token and obtain a CA signing token. Enforce authentication/authorization on this route (or at the ingress) before merging.

rustyjux: this is the pattern across the board for the provisioner API - it's internal only. 
OpenAPI showing `jwt` might be misleading or maybe this is reasonable because that is expected in terms of calls to the portal which then calls out with the provisioner service?

provisioner-api/src/controllers/runtime-groups-controller.ts
ikethecoder: True, will refactor it so the "san" and "subject" are not passed in the body, and those get calculated on the provisioner side before calling the ca-token api.

provisioner-api/src/routes/runtime-groups.ts
ikethecoder: True, will remove the `jwt` security.

PR #1523, opened 2026-08-05
SP136: subsystem RBAC roles + gateway-scoped delegation for oas-services & connections

ikethecoder:
## Summary

Started as exposing `GET/PUT /organizations/{org}/subsystems/{name}/access` (subsystem RBAC role assignment); grew into wiring those roles up to real permissions and adding a gateway-scoped delegation path as an alternative to org-wide `System.Manage`.

- **`GET/PUT /organizations/{org}/subsystems/{name}/access`** — view/change a subsystem's RBAC role assignments (unchanged from the original PR).
- **Renamed** the subsystem role `system-owner` → `subsystem-owner`, and the org-level `System.Manage`-derived role `system-owner` → `system-admin` (`scopesToRoles()`, `matrix.csv`) — these two previously shared an ambiguous name across different scopes.
- **Made the three subsystem/runtime-group roles functional.** `subsystem-owner`, `tech-lead`, and `access-manager` previously carried an empty, purely informational permission set. They now grant real, enforced UMA2 permissions on the gateway's namespace resource: `subsystem-owner` gets `Namespace.Manage` + `Subsystem.Manage` + `GatewayPattern.Publish`, `tech-lead` gets `Subsystem.Manage` + `GatewayPattern.Publish`, `access-manager` gets `Connection.Manage`. This role system now also applies to **runtime groups** (`registerRuntimeGroupGateway`), not just subsystems — though only `tech-lead`/`subsystem-owner` are auto-assigned there, not `access-manager`.
- **New `Subsystem.Manage` scope** — a gateway-scoped alternative to `System.Manage` for `oas-services` (create/list/get/spec/delete) and `connections` (create/delete), stacked additively so existing `system-admin` holders are unaffected. Resolves to the specific service/subsystem gateway per-operation; falls back to a UMA2 discovery-based check for listing (a `Subsystem.Manage`-only caller sees only their own gateways' services, not the whole org, and not a 403).
- **`Connection.Manage` can now also list connections**, filtered the same discovery-based way.
- **Catalog service-detail endpoint is now public**, matching the already-public spec endpoint.
- **`docs/sdx-v1-permissions-audit.md`** — new doc auditing every sdx/v1 endpoint's required scope and what resource it's actually checked against, plus a section on a residual inconsistency: `Subsystem.Manage` is gateway-scoped via the sdx/v1 REST layer but unfiltered (org-wide) via direct Keystone GraphQL access, mirroring how `System.Manage` already behaved.

## Test plan
- [x] `tsc --noEmit` clean.
- [x] `npm run tsoa-build-sdx-v1` regenerates cleanly.
- [x] Full backend jest suite green (243 passing; the only failures are pre-existing `integrated-*` tests that require a live Keycloak/MongoDB connection unavailable in CI/sandbox).
- [x] New cypress spec `e2e/cypress/tests/99-sp136/01-subsystem-access.ts` (from the original PR) covers the role-assignment endpoints; updated for the `subsystem-owner` rename.
- [x] `e2e/cypress/tests/19-api-v3/02-organization.ts` updated for the new role permission sets returned by `GET /roles`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rustyjux (approved):
Looks good. 
Couple comments to consider:
1. No activity logging, not sure when this will be implemented
2. Probably just a weird edge case, but any chance someone would configure subsystem access before registering the gateway (which would overwrite the configured access)?

src/controllers/sdx/v1/OrgSubsystemController.ts
```diff
@@ -184,4 +188,103 @@ export class OrgSubsystemController extends Controller {
...
+   * @summary Update subsystem RBAC access
+   *
+   * @param org - Organization identifier
+   * @param name - Subsystem name
+   * @param body - Desired RBAC role membership for the subsystem
+   * @param request - HTTP request object for context creation
+   */
+  @Put('/{name}/access')
+  @OperationId('putSubsystemAccess')
+  @Security('jwt', ['System.Manage'])
```
phowells, line 255: Could you please confirm the intended authorization boundary for changing subsystem membership? The existing organization access `PUT` uses `GroupAccess.Manage`, while this RBAC mutation uses `System.Manage`. Is `System.Manage` intentionally sufficient to grant or revoke the `system-owner`, `tech-lead`, and `access-manager` roles?

src/controllers/sdx/v1/OrgSubsystemController.ts
```diff
@@ -184,4 +188,103 @@ export class OrgSubsystemController extends Controller {
...
+  public async putAccess(
+    @Path() org: string,
+    @Path() name: string,
+    @Body() body: { members: GroupMember[] },
```
phowells, line 259: Could the accepted role names be validated before this full-membership sync? `GroupMember.roles` is a `string[]`, while `SysGroupAccessService` only processes the predefined system roles. An unrecognized or misspelled role is therefore ignored while absent recognized roles are synchronized as empty, which can revoke current assignments. Would rejecting unknown role names protect against accidental destructive updates?

src/controllers/sdx/v1/OrgSubsystemController.ts
```diff
@@ -184,4 +188,103 @@ export class OrgSubsystemController extends Controller {
...
+    const prodEnv = await getGwaProductEnvironment(this.keystone.sudo(), false);
+    const envConfig = prodEnv.issuerEnvConfig;
+
+    const sysGroupAccessService = new SysGroupAccessService(prodEnv.uma2);
+    await sysGroupAccessService.login(
+      envConfig.clientId,
+      envConfig.clientSecret
+    );
+
+    await sysGroupAccessService.createOrUpdateGroupAccess(
```
phowells, line 280: Could you clarify whether subsystem access changes should produce activity or audit records? The existing organization-access `PUT` uses the returned granted/revoked diff to call `logOrganizationAccessChanges`, while the diff from this RBAC update is discarded. If that omission is intentional, some context here would be helpful.

Elson9: `createOrUpdateGroupAccess` performs a full membership sync, but silently drops emails that cannot be resolved in Keycloak. As a result, a request containing an unknown email returns `204` while revoking the existing members for all three roles.

Could we validate that every requested member resolves before starting the sync, and return `400` if any do not?

PR #1524, opened 2026-08-05
SP136: Return RBAC roles via includeAccess on /catalog/organizations

ikethecoder:
## Summary
- `GET /catalog/organizations` gains an opt-in `includeAccess` query param, mirroring the existing convention on `GET /catalog/subsystems/{name}` (`CatalogController.getSubsystem`).
- When `includeAccess=true`, each organization entry is enriched with an `access: GroupMember[]` field, sourced from `GroupAccessService.getGroupMembership(org.name)` — the same primitive `GET /organizations/{org}/access` already uses — rather than the subsystem-scoped `EnrichWithAccess`, which doesn't apply to organizations.
- Default behavior (`includeAccess` omitted) is unchanged.

Plan: `new_feedback/SP136-catalog-organizations-include-access.md`.

## Test plan
- [x] New e2e spec `e2e/cypress/tests/99-sp136/02-catalog-organizations-access.ts`, confirmed failing against unmodified `hotfix/aps-4799-m0-carryover` (missing `access` field), then passing after the fix.
- [x] Added `./cypress/tests/99-*/*.ts` to `e2e/cypress.config.ts` `specPattern` (not yet present on this base branch).
- [x] Regression: `e2e/cypress/tests/19-api-v3/02-organization.ts` (existing `GET/PUT /organizations/{org}/access`) — all passing.
- [x] Regression: `e2e/cypress/tests/21-sdx-api/v1/01-subsystems.ts` — all passing.
- [x] Deployed via the `deploy-branch` skill (rebuilt/restarted `apsportal` + `provisioner`) and verified locally against the docker-compose stack.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/controllers/sdx/v1/CatalogController.ts
+++ b/src/controllers/sdx/v1/CatalogController.ts
@@ -43,6 +43,8 @@ import {
 import { KeystoneService } from '../../ioc/keystoneInjector';
 import assert from 'assert';
 import { ResourceScope } from '../../../services/workflow/openapi-spec-loader';
+import { GroupAccessService } from '../../../services/org-groups';
+import { getGwaProductEnvironment } from '../../../services/workflow';
 
 interface MissingCredentialsJSON {
   code: 'credentials_required' | 'invalid_token';
@@ -94,9 +96,12 @@ export class CatalogController extends Controller {
 
   @Get('/organizations')
   @OperationId('organization-list')
-  public async listOrganizations(): Promise<any[]> {
-    const orgs = await getOrganizations(this.keystone.sudo());
-    return orgs
+  public async listOrganizations(
+    @Query('includeAccess') includeAccess: boolean = false
+  ): Promise<any[]> {
+    const ctx = this.keystone.sudo();
+    const orgs = await getOrganizations(ctx);
+    const result: any[] = orgs
       .map((o) => ({
         name: o.name,
         title: o.title,
@@ -104,8 +109,25 @@ export class CatalogController extends Controller {
         publicBodyId: o.publicBodyId,
         member: getOrganizationMemberDetails(o.tags),
       }))
-      .filter((o) => o.member !== undefined)
-      .map((o) => removeEmpty(o));
+      .filter((o) => o.member !== undefined);
+
+    if (includeAccess) {
+      const prodEnv = await getGwaProductEnvironment(ctx, false);
+      const groupAccessService = new GroupAccessService(prodEnv.uma2);
+      await groupAccessService.login(
+        prodEnv.issuerEnvConfig.clientId!,
+        prodEnv.issuerEnvConfig.clientSecret!
+      );
+
+      for (const org of result) {
+        const membership = await groupAccessService.getGroupMembership(
+          org.name
+        );
+        org.access = membership?.members ?? [];
+      }
+    }
+
+    return result.map((o) => removeEmpty(o));
   }
 
   /**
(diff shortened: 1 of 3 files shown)


rustyjux (approved):
Biggest concern is the calls to Keycloak by `getGroupMembership` introducing latency with a high number of orgs. As is, per org, `getGroupMembership` does roughly:

- one Keycloak group search by name
- one list members call per matching role path (where we might expect 2: organization-admin + system-admin)
- 
So ~3 Keycloak round-trips per org done sequentially in a for/await loop. Maybe better would be concurrent using `Promise.all`?

Also, to confirm - the intention is the public catalog will return emails of folks w/ org roles?

src/controllers/sdx/v1/CatalogController.ts
```diff
@@ -94,18 +96,38 @@ export class CatalogController extends Controller {
...
         title: o.title,
         description: o.description,
         publicBodyId: o.publicBodyId,
         member: getOrganizationMemberDetails(o.tags),
       }))
-      .filter((o) => o.member !== undefined)
-      .map((o) => removeEmpty(o));
+      .filter((o) => o.member !== undefined);
+
+    if (includeAccess) {
```
phowells, line 138: Could you please clarify whether access membership is intended to be public here? This catalog route has no `@Security` decorator, so `?includeAccess=true` allows an unauthenticated caller to receive members and their RBAC roles for every organization. If that is not intended, should the access-enriched variant require authorization?

src/controllers/sdx/v1/CatalogController.ts
```diff
@@ -94,18 +96,38 @@ export class CatalogController extends Controller {
...
+      const prodEnv = await getGwaProductEnvironment(ctx, false);
+      const groupAccessService = new GroupAccessService(prodEnv.uma2);
+      await groupAccessService.login(
+        prodEnv.issuerEnvConfig.clientId!,
+        prodEnv.issuerEnvConfig.clientSecret!
+      );
+
+      for (const org of result) {
```
phowells: Could these membership lookups be batched or parallelized? This adds one sequential Keycloak request per catalog organization, so `includeAccess=true` latency grows as the sum of every lookup and a single lookup failure aborts the full listing. Is the expected organization count small enough that this is deliberate?

src/controllers/sdx/v1/CatalogController.ts
ikethecoder: yes, it is intended to be public information - contact details.

src/controllers/sdx/v1/CatalogController.ts
ikethecoder: True, will have a look at that and see what options are available - unfortunately Keycloak has a limited search capability (i.e./ no: name_in() capability), but some parallel action might help.

src/controllers/sdx/v1/CatalogController.ts
```diff
@@ -94,18 +96,38 @@ export class CatalogController extends Controller {
...
+        const membership = await groupAccessService.getGroupMembership(
+          org.name
+        );
+        org.access = membership?.members ?? [];
```
Elson9: `getGroupMembership()` includes Keycloak `id`, `username`, and display-name fields. Since this public catalog endpoint returns those objects directly, could we sanitize the memberships first (as `getSubsystemRoles()` does) to expose only the intended contact details and roles?

src/controllers/sdx/v1/CatalogController.ts
ikethecoder: Perhaps I do have to review before handling it over for PR lol.  The "includeAccess" is meant for a GET operation that returns details for a particular organization.  So by moving it off of the list org and onto this new operation, the performance concern goes away.

PR #1525, opened 2026-08-05
APS-4799 m0 carryover

ikethecoder:
## Summary

M0 carryover work for APS-4799 (SP136), bringing three previously merged pieces into this hotfix branch:

- **Subsystem RBAC roles & gateway-scoped delegation** — adds RBAC role support for subsystems and gateway-scoped delegation for OAS services and connections, including a new subsystem controller/service layer and matrix permission updates.
- **RBAC roles on `/catalog/organizations`** — the `includeAccess` option on the catalog organizations endpoint now returns RBAC roles.
- **Cert-signing tokens routed through the provisioner** — runtime group cert-signing token issuance now goes through the provisioner API via a new CA token client/service, replacing the old direct step-token integration.

Also includes minor CI/build workflow updates (build-only gating, buildx/qemu fixes) and a permissions audit doc.

## Test plan
- [ ] Existing e2e suites for subsystem access, catalog organizations access, and cert-token provisioner routing (`e2e/cypress/tests/99-sp136/*`) pass
- [ ] Unit tests under `src/test/services/org-groups/*` pass
- [ ] Manual verification of RBAC role visibility on `/catalog/organizations` and runtime group cert-signing flow via the provisioner


src/controllers/sdx/v1/OrgConnectionController.ts
```diff
@@ -95,18 +100,50 @@ export class OrgConnectionController extends Controller {
...
     @Request() request: any
   ): Promise<ConnectionRequest[]> {
     const ctx = this.keystone.createContext(request);
-    const records = await new ConnectionService().listConnectionsByOrganization(
-      ctx,
-      org
-    );
+    const connectionService = new ConnectionService();
+
+    const callerScopes = (request.oauth_user?.scope || '').split(' ');
```
rustyjux: This still keys off the raw JWT (`oauth_user.scope`) rather than the scope that actually authorized the request (`request.user.scope` after #1528’s sequential OR).

If someone only gets in via `Connection.Manage` but their token still mentions `System.Manage` (e.g. stale after demotion), `hasOrgWideAccess` is true and we return org-wide results.

Suggest: `request.user?.scope === 'System.Manage'` (and apply same pattern in the OAS services list handler).

rustyjux (approved):
Looks good overall — happy to approve.

One possible follow-up noted re authz for list endpoints.

PR #1526, opened 2026-08-05
SP136: Subsystem gateway registration is create-only, no partial-state recovery

ikethecoder:
## Summary

`PUT /organizations/{org}/subsystems/{name}/gateway` always targets a subsystem's fixed, pre-allocated namespace ID, but the registration flow (`CreateNamespaceForSubsystem` → `createSDXNamespace` → `CreateNamespace`) was create-only: `checkNamespaceAvailable` hard-rejected any second call for that namespace, and `createResourceSet`/`createUmaPolicy` had no idempotency. A registration interrupted partway (timeout, pod restart, dropped connection) could never be retried or completed — every subsequent call just failed with `Namespace already exists`, forever.

- `src/services/org-groups/namespace.ts`: add `namespaceExists`, a non-throwing check alongside the existing `checkNamespaceAvailable`.
- `src/services/workflow/create-namespace.ts`: add `CreateNamespaceArgs.allowResume`. When set and the namespace already exists, skip the hard failure and instead reuse the existing UMA resource set (`findResourceByName`), upsert permission tickets (`createOrUpdatePermission`), and always reconcile the `ns` group's attributes instead of only on first create. Off by default — every other namespace-creation caller (org, runtime group) is unaffected.
- `src/services/workflow/create-namespace-sdx.ts`: `createSDXNamespace` (the only caller with a stable, pre-allocated namespace ID) passes `allowResume: true` and calls the new `createUmaPolicyIfMissing`.
- `src/services/workflow/ns-uma-policy-access.ts`: add `createUmaPolicyIfMissing`, which skips creating a duplicate UMA policy for a client that already has one on the resource.

No `provisioner-api` changes — gateway registration never calls the provisioner; it only grants the `sdx-provisioner` service account UMA scopes for its own later, already-idempotent publish flow.

## Test plan

- [x] New e2e test `e2e/cypress/tests/99-sp136/01-subsystem-gateway-registration-recovery.ts`: registers a subsystem's gateway, then registers it again (simulating a retry) and asserts `200` + the same `gatewayId`, then confirms the subsystem list is still consistent.
  - Confirmed this fails against the pre-fix code (`422 Namespace already exists`).
  - Confirmed it passes after the fix.
- [x] `e2e/cypress/tests/21-sdx-api/v1/01-subsystems.ts` (existing subsystem CRUD/gateway suite) — 11/11 passing, no regression.

### DIFF
--- a/e2e/cypress.config.ts
+++ b/e2e/cypress.config.ts
@@ -46,6 +46,7 @@ export default defineConfig({
         './cypress/tests/21-*/**/*.ts',
         './cypress/tests/22-*/*.ts',
         './cypress/tests/23-*/*.ts',
+        './cypress/tests/99-*/*.ts',
       ]
       return config
     },
--- a/e2e/cypress/tests/99-sp136/01-subsystem-gateway-registration-recovery.ts
+++ b/e2e/cypress/tests/99-sp136/01-subsystem-gateway-registration-recovery.ts
@@ -1,0 +1,63 @@
+import { v4 as uuidv4 } from 'uuid'
+
+import { createRuntimeGroup, createSubsystem, uniqueSubsystemName } from '../../support/sdx-commands'
+
+describe('SDX Subsystem Gateway Registration Recovery', () => {
+  let workingData: any
+
+  before(() => {
+    cy.buildOrgGatewayDatasetAndProduct().then((data) => {
+      workingData = data
+
+      const rg = uuidv4().replace(/-/g, '').toUpperCase().substring(0, 6)
+      workingData['runtimeGroupId'] = rg.toLowerCase()
+
+      createRuntimeGroup(workingData.org, workingData.runtimeGroupId, 'dev')
+    })
+  })
+
+  it('PUT /organizations/{org}/subsystems/{name}/gateway - retrying a registration for the same subsystem recovers instead of failing', () => {
+    const { org, runtimeGroupId } = workingData
+    const subsystemName = uniqueSubsystemName()
+
+    createSubsystem(org, subsystemName, () => {
+      cy.setRequestBody({ runtimeGroupName: runtimeGroupId })
+      cy.callAPI(
+        `ds/api/sdx/v1/organizations/${org.name}/subsystems/${subsystemName}/gateway`,
+        'PUT'
+      ).then(({ apiRes: { status, body } }: any) => {
+        expect(status, body.message).to.be.equal(200)
+        expect(body).to.have.property('gatewayId')
+        const gatewayId = body.gatewayId
+
+        // Simulate a caller retrying the same registration request - e.g.
+        // because it never saw the first response (timeout, dropped
+        // connection, portal restart). The subsystem's gateway/namespace ID
+        // is fixed and immutable, so this always targets the exact same
+        // namespace as the first call.
+        cy.setRequestBody({ runtimeGroupName: runtimeGroupId })
+        cy.callAPI(
+          `ds/api/sdx/v1/organizations/${org.name}/subsystems/${subsystemName}/gateway`,
+          'PUT'
+        ).then(({ apiRes: { status, body } }: any) => {
+          expect(status, body.message).to.be.equal(200)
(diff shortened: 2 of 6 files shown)


e2e/cypress/tests/99-sp136/01-subsystem-gateway-registration-recovery.ts
```diff
@@ -0,0 +1,63 @@
...
+  let workingData: any
+
+  before(() => {
+    cy.buildOrgGatewayDatasetAndProduct().then((data) => {
+      workingData = data
+
+      const rg = uuidv4().replace(/-/g, '').toUpperCase().substring(0, 6)
+      workingData['runtimeGroupId'] = rg.toLowerCase()
+
+      createRuntimeGroup(workingData.org, workingData.runtimeGroupId, 'dev')
```
rustyjux, line 15: any way to predictably simulate a failed run to create partial state?

src/services/workflow/create-namespace-sdx.ts
```diff
@@ -296,6 +299,7 @@ async function createSDXNamespace(
     'GatewayPattern.Publish',
   ];
   args.includeSDXScopes = true;
+  args.allowResume = true;
```
rustyjux, line 302: PR only talks about subsystem, but this will apply to org and runtime group gateways too. Those all have fixed gateway names so should be fine too.

src/services/workflow/create-namespace.ts
```diff
@@ -67,7 +75,15 @@ export async function CreateNamespace(
     envCtx.issuerEnvConfig.clientId,
     envCtx.issuerEnvConfig.clientSecret
   );
-  await nsService.checkNamespaceAvailable(newNS);
+  const resuming = args.allowResume && (await nsService.namespaceExists(newNS));
```
rustyjux, line 78: **Resume is keyed only on the Keycloak `ns` group**  
If the first attempt created the **UMA resource** but died **before** the `ns` group, retry thinks it’s a fresh create and may fail on `createResourceSet` (name already taken). The stuck cases this PR fixes well are “group exists, rest incomplete.” The earlier half of the pipeline is still fragile.

Fix would involve treating 'already in progress' as the group _or_ UMA resource already exists.

rustyjux (changes requested):
I think we should cover the case of UMA resource exists but group doesn't - see comment on create-namespace.ts

e2e/cypress.config.ts
```diff
@@ -46,6 +46,7 @@ export default defineConfig({
         './cypress/tests/21-*/**/*.ts',
         './cypress/tests/22-*/*.ts',
         './cypress/tests/23-*/*.ts',
+        './cypress/tests/99-*/*.ts',
```
phowells, line 49: Could the existing duplicate-registration cases be updated as part of this change? `21-sdx-api/v1/05-gateways.ts` is already included above and still expects the second runtime-group and subsystem gateway PUTs to return 422 with `Namespace already exists`, while the new `99-sp136` case expects the second PUT to return 200. With this pattern enabled, the full E2E suite has contradictory expectations and the older cases appear likely to fail.

src/services/workflow/create-namespace.ts
```diff
@@ -128,9 +147,13 @@ export async function CreateNamespace(
     envCtx.issuerEnvConfig.clientSecret
   );
 
-  const { id, created } = await kcGroupService.createIfMissing('ns', newNS);
+  const { id } = await kcGroupService.createIfMissing('ns', newNS);
 
-  if (created) {
+  {
+    // Reconcile attributes unconditionally (not just on first create) so a
```
phowells, line 153: Could we guard the resume path against a different requested configuration? A second subsystem gateway PUT can supply another `runtimeGroupName`; once resume is allowed, this block overwrites `perm-runtime-group` and rewrites its domains. That appears to turn the create-only endpoint into a gateway move/update rather than an idempotent retry. If moving is intended, could that behavior be documented and tested? Otherwise, could the resume path verify that the stored assignment matches the request and reject a mismatch?

PR #1527, opened 2026-08-06
Feature/aps 4674 self issue cred

Elson9:
## Summary
- Adds gateway-scoped REST APIs under `/ds/api/v3` for API Providers to **issue** and **regenerate** consumer credentials with UMA scope `CredentialIssuer.Generate` (no Access Request / approval queue).
- Supports `kong-api-key-only`, `kong-api-key-acl`, and `client-credentials` (client-secret, client-jwt, JWKS URL/cert), including application reuse across environments, optional labels, and ownerless issuer-created Applications excluded from My Access.
- Adds independent local Cypress suite `24-self-issuing-credentials` covering issue/reuse/duplicate/regenerate/authz/upstream use, Consumers UI label filter and revoke, and My Access exclusion for ownerless apps; fixes JWT regenerate to update certificates via the same path as issue.

## Test plan
- [x] Unit: issue/regenerate workflow and `CredentialIssuer.Generate` scope-role tests
- [x] Local Cypress: `npx cypress run --config-file cypress.config.ts --spec 'cypress/tests/24-self-issuing-credentials/*.cy.ts'`
- [x] DEV smoke: SA with `CredentialIssuer.Generate` → POST issue → Consumers UI + labels → reuse app on second env → duplicate fails → PUT regenerate → 403 without scope → UI revoke
- [x] Confirm JWT regenerate returns new key material; JWKS-URL regenerate remains unsupported
- [x] Confirm My Access does not list ownerless issuer-created applications (`04-my-access-exclusion.cy.ts`)

---
🚀 Feature branch deployment: https://api-services-portal-feature-aps-4674-self-issue-cred.apps.silver.devops.gov.bc.ca


src/controllers/v3/GatewayConsumersController.ts
```diff
@@ -0,0 +1,109 @@
...
+    const ctx = this.keystone.createContext(request, true);
+    const credential = await issueGatewayCredential(ctx, gatewayId, {
+      environmentAppId: body.environmentAppId,
+      application: body.application || {},
+      labels: body.labels,
+      controls: body.controls as any,
+    });
+
+    this.setStatus(201);
+    return credential;
```
ikethecoder: There is a `removeEmpty()` function that we use in other controllers to remove any `null` values.  Perhaps could do for this, otherwise the result provides irrelevant data.  For example:
```json
{
  "flow": "client-credentials",
  "clientId": "EA7F0EF6-F3FBBBC03FF",
  "clientSecret": "***",
  "issuer": null,
  "tokenEndpoint": "http://keycloak.localtest.me:9081/auth/realms/master/protocol/openid-connect/token",
  "clientPublicKey": null,
  "clientPrivateKey": null
}
```

ikethecoder (commented):
Perhaps a bit of decoration on this - maybe adding the Application description (?) or for "Application Owner" have some gray-italic content around: "Application administered by this gateway" (?)

Image

ikethecoder (commented):
The "Edit" button that brings up the dialog seems to be unhappy:

Image

ikethecoder (commented):
Deleting a consumer leaves the application as an orphan.  May need to clean up the application when deleting the consumer.

ikethecoder (changes requested):

ikethecoder:
Was able to get a credential - few suggestions above.  As far as places that require the most help, it wasn't clear where to find the `environmentAppId`, as it does not appear on the Portal, and the other is the `controls {}` - when to use them, when not to, etc.  I haven't done a code review yet, but functionally works great!

src/services/workflow/issue-gateway-credential.ts
```diff
@@ -0,0 +1,461 @@
...
+    clientId,
+    false
+  );
+  assert.strictEqual(
+    typeof existingConsumer === 'undefined',
+    true,
+    'This application already has access to this environment'
+  );
+
+  const { newCredential, serviceAccessId, consumer } = await createCredential(
```
phowells: Could we extend the rollback boundary to include `createCredential` (and the new `Application`, when applicable)? This call performs several external and Keystone writes before a `ServiceAccess` ID is returned. For example, if `forceSync` or `addServiceAccess` fails after the Kong key/client or Keycloak client is created, the `catch` below is never entered, leaving partial resources behind and making a retry likely to fail. The current cleanup tests only cover failures after this call returns.

src/services/workflow/issue-gateway-credential.ts
```diff
@@ -0,0 +1,461 @@
...
+    false,
+    'Initial Access Token is required when doing client registration via an IAT'
+  );
+}
+
+async function validateControls(
+  controls: RequestControls,
+  productEnvironment: Environment
+) {
+  if (controls.jwksUrl) {
```
phowells: Could we restrict `jwksUrl` before making this server-side request—for example to HTTPS and non-loopback, non-link-local, and non-private destinations, with redirects checked as well? `IsJWKSURLValid` follows a caller-controlled URL and currently accepts any reachable JSON response, so a token with `CredentialIssuer.Generate` could use this endpoint to probe internal HTTP services.

src/services/workflow/issue-gateway-credential.ts
```diff
@@ -0,0 +1,461 @@
...
+    );
+  } else if (controls.clientCertificate) {
+    assert.strictEqual(
+      IsCertificateValid(controls.clientCertificate),
+      true,
+      'Certificate failed validation'
+    );
+  }
+
+  if (
```
phowells: Could we enforce a valid signing-control combination for the configured authenticator before creating the client? As currently modelled, all three fields are optional and combinable: a `client-jwt` request without a supplied or generated certificate returns no secret or private key, while `jwksUrl` plus `clientGenCertificate` returns a generated key even though Keycloak is configured to use the URL. Both can yield credentials that cannot authenticate. Also, `lookupProductEnvironmentServices` does not select `clientAuthenticator`, so this branch cannot currently inspect it.

src/controllers/v3/types-extra.ts
```diff
@@ -19,3 +19,82 @@ export interface PublishResult {
...
+  name?: string;
+  description?: string;
+}
+
+/**
+ * @tsoaModel
+ */
+export interface IssueGatewayConsumerControls {
+  defaultClientScopes?: string[];
+  defaultOptionalScopes?: string[];
```
phowells, line 61: Could we either apply `defaultOptionalScopes` or remove it from this new request contract? `issueGatewayCredential` forwards the field, but `setupAuthorizationAndEnable` calls `syncAndApply(clientId, clientScopes, [])`, so a caller can receive 201 even though the requested optional scopes were silently ignored.

phowells (changes requested):

src/services/workflow/issue-gateway-credential.ts
Elson9: Thanks, this is a valid security concern. However, `IsJWKSURLValid` is shared by the existing Access Request and credential update flows, so changing it here could affect existing production clients using HTTP or private JWKS endpoints. I’ve reverted the hardening from this PR to keep the ticket scoped to the new Credential Issuer API. This should be addressed separately with an audit of existing JWKS URLs and a planned rollout strategy.

Elson9: Created APS-4836 for this

src/services/workflow/issue-gateway-credential.ts
```diff
@@ -0,0 +1,683 @@
...
+  const nickname = clientId;
+
+  let serviceAccessId: string | undefined;
+  let kongConsumerExtId: string | undefined;
+  let keystoneConsumerId: string | undefined;
+  let keycloakClientCreated = false;
+
+  try {
+    if (flow == 'kong-api-key-acl' || flow == 'kong-api-key-only') {
+      const newApiKey = await registerApiKey(
```
phowells, line 443: One rollback path still seems to fall outside the new tracking. `registerApiKey` creates the Kong consumer first, then adds the key and Keystone record before returning. If either later await rejects, this assignment never completes, so `kongConsumerExtId` and `keystoneConsumerId` remain undefined and `cleanupPartialCredentialResources` cannot remove what was already created. Can we move the rollback into `registerApiKey` (or expose progress as each resource is created) so failures inside the helper are covered too? A test where `addKeyAuthToConsumer` or `addKongConsumer` rejects would pin this down.

Update: @Elson9, I added one possible implementation in [commit `284aa8d0`](https://github.com/bcgov/api-services-portal/commit/284aa8d02b2048fa385a333846954fa1fb6fb1f5), which is now on this PR branch. Please adjust it if another approach fits better.

src/services/workflow/issue-gateway-credential.ts
```diff
@@ -0,0 +1,683 @@
...
+      keycloakClientCreated = true;
+
+      const kongApi = new KongConsumerService(process.env.KONG_URL);
+      const kongConsumer = await kongApi.createKongConsumer(
+        nickname,
+        clientId,
+        application
+      );
+      kongConsumerExtId = kongConsumer.id;
+      keystoneConsumerId = await AddClientConsumer(
```
phowells, line 522: There is a similar gap here: `AddClientConsumer` calls `addKongConsumer` and then `forceSync`, returning the PK only after both finish. If its internal `forceSync` fails, the GatewayConsumer exists but this assignment never sets `keystoneConsumerId`; cleanup removes the Kong and Keycloak resources but leaves that Keystone record behind. It also makes the `forceSync` below a second sync of the same consumer. Would it make sense to have `AddClientConsumer` clean up its partial write (and drop the redundant sync), or split the steps here so the PK is recorded before syncing?

Update: [Commit `284aa8d0`](https://github.com/bcgov/api-services-portal/commit/284aa8d02b2048fa385a333846954fa1fb6fb1f5) is now on this PR branch and covers this path, including removing the duplicate sync.

phowells (approved):

src/authz/graphql-whitelist/httplocalhost4180managerconsumers-b2cc08.gql
```diff
@@ -0,0 +1,51 @@
+
```
ikethecoder: Rather than creating a new whitelist gql - can you find the existing one update it to what this one is?  So its clear that its updating an existing graphql query rather than introducing a new one

ikethecoder (approved):
James, all works great.  I don't think we have to hold up the feature, but we should get a new ticket that deals with the activity logging - atm nothing shows up in the activity log regarding creating/regenerating.

PR #1544, opened 2026-08-14
APS-4790 Update connection requests with provisioning status

phowells:
## Summary

APS-4790 updates each Connection Request with the outcome of provisioning:

- `pending` when apply provisioning begins
- `provisioned` when every dispatched provider batch succeeds
- `failed` when a provider batch fails or is skipped, or when provisioning throws
- an optional, safe status message with useful context

On successful provisioning, data emitted by `Information` resources is also copied into the `provisionerStatus` payload. This makes connection details such as the provisioned `endpoint` and relative `spec` path directly available to clients while preserving arbitrary Information fields that patterns may add later. The controlled `status` and `message` values cannot be overwritten by Information data.

Provisioner status writes use the portal's existing internal feed/batch path. This avoids granting the single `sdx-provisioner` client a new UMA permission or mapping a client ID directly to a Keystone role.

Preview, diff, and delete operations do not change the provisioning status.

### 1. `354459c4` — Prevent provisioner status callbacks

Adds a focused Connection Request change detector and updates the Keystone `afterChange` hook to call the provisioner only when provisioning-relevant fields changed.

A feed update containing only `provisionerStatus` therefore persists the result without starting another provisioning cycle. Tests cover creates, status-only updates, repeated identifiers, and real provisioning-field changes.

### 2. `3a5020cf` — Record connection provisioning status

Extends the provisioner's Connection Request types and adds a typed connection-status service backed by `PUT /feed/ConnectionRequest`.

The connection controller now:

- records `pending` before apply processing
- records `provisioned` after complete success
- records `failed` for partial failures, skipped provider batches, and thrown exceptions
- uses safe messages that do not expose upstream exception details
- leaves preview, diff, and delete status unchanged

It also excludes the persisted status from Cedar policy input and adds focused controller and service tests.

### 3. `8a6f60cd` — Add connection status datafix

Adds a noninteractive `npm run datafix:aps-4790` command to the provisioner image.

The datafix:

- retrieves Connection Requests through a dedicated, paginated internal feed endpoint
- deduplicates records
- selects only active requests without a recognized provisioning status
- reprovisions candidates sequentially through the normal connection controller
- continues after individual failures
- logs a final summary and exits nonzero when any candidate fails

The command is idempotent for requests already marked `pending`, `provisioned`, or `failed`. A separate one-time post-deployment Job in `api-serv-infra` will execute it after the portal and provisioner images are deployed.

### 4. `64a3296d` — Include provisioning information in status

Collects the `details` produced by successfully applied `Information` resource batches and copies their fields into the final `provisioned` status. Multiple Information results are merged in dispatch order, and `status` and `message` remain authoritative.

The status contracts accept future Information fields without requiring a database or Keystone model change. The normal apply path is shared by the one-time datafix, so existing Connection Requests receive the same endpoint/spec information when they are reprovisioned.

Focused tests cover Information aggregation, protection of the controlled status fields, and persistence of arbitrary Information fields through the feed JSON representation.

## Validation

- Provisioner TypeScript typecheck passed.
- Provisioner build passed.
- All provisioner test files passed, including Information propagation and controlled-field protection.
- 12 focused portal tests passed, covering recursion prevention, feed enumeration and JSON persistence, and existing connection-service behavior.
- `git diff --check` passed.
- The full portal TypeScript build still reports the existing Keycloak client typing errors in `services/keycloak/group-service.ts`; APS-4790 introduces no additional TypeScript errors.

---
🚀 Feature branch deployment: https://api-services-portal-feature-aps-4790-prov-status.apps.silver.devops.gov.bc.ca


src/batch/feed-worker.ts
```diff
@@ -312,6 +312,38 @@ export const getRecords = async function (
   );
 };
 
+export const listConnectionRequestsFeedWorker = async (
```
ikethecoder: Would like to keep entity-specific functions out of the `feed-worker`.  I think either leveraging getRecords or creating a `listFeedWorker` should be fine.  The JSON string storage is a bit unfortunate.. the `data-rules` defines transformations which could be used to create the reciprocal transformation.  Alt, you could add a filter on `/catalog/organizations` to filter out orgs that have connections - or just use the whole list.  And then use that to call `/organizations/{org}/connections` for each org.  It would mean one call per org, but for a one-off datafix, probably ok.

ikethecoder (changes requested):
Like the datafixes!  Think that pattern will get used in future when we may have to bulk-change the kong configurations.

src/server.ts
```diff
@@ -321,6 +322,13 @@ const configureExpress = (app: any) => {
       res.status(400).json({ result: 'error', error: '' + err });
     });
   });
+  app.get('/feed/ConnectionRequest', (req: any, res: any) => {
```
Elson9: `/feed` is on the oauth2-proxy skip-auth list, and this handler uses `skipAccessControl`. That makes `GET /feed/ConnectionRequest` a public dump of every connection, including `requesterDetails`, `clientResources`, and `serviceResources`. Existing feed GETs only return `extForeignKey` / `extRecordHash`.

Aidan's suggestion to use `/catalog/organizations` and `/organizations/{org}/connections` would keep this on the authenticated SDX API.

provisioner-api/src/datafixes/aps-4790.ts
```diff
@@ -0,0 +1,128 @@
...
+}
+
+export interface Aps4790DatafixSummary {
+  connections: number;
+  candidates: number;
+  provisioned: number;
+  failed: number;
+}
+
+const PROVISIONER_STATUSES = new Set(['pending', 'provisioned', 'failed']);
```
Elson9: Treating `pending` as already complete means a killed Job, or a successful apply whose final status PUT failed, will be skipped on retry and stay `pending` forever.

For this one-shot datafix, I'd only skip `provisioned` and `failed`. `pending` should still be a candidate.

provisioner-api/src/controllers/connections-controller.ts
```diff
@@ -98,6 +112,81 @@ export class ConnectionsController {
...
+          .join(', ');
+        await this.updateProvisionerStatus(
+          connectionRequest,
+          'failed',
+          `Provisioning failed for ${unsuccessful} provider batch${
+            unsuccessful === 1 ? '' : 'es'
+          }${providers ? `: ${providers}` : ''}.`
+        );
+      } else {
+        await this.updateProvisionerStatus(
```
Elson9, line 131: `pending` is recorded before the try, but this final `provisioned` / `failed` write sits outside it. If apply succeeds and this PUT throws, the connection stays `pending` even though resources were applied — and the datafix will not retry it.

You could wrap this the same way as `updateFailedStatusAfterException`, so a status-write failure can't leave a successful provision stuck at `pending`.

provisioner-api/src/controllers/connections-controller.test.ts
```diff
@@ -0,0 +1,223 @@
...
+    await controller.onConnectionRequestChange(
+      'connection-1',
+      connectionRequest,
+      action
+    );
+    assert.deepEqual(statusUpdates, []);
+  }
+});
+
+test('delete does not update provisioner status', async () => {
```
Elson9, line 215: Delete/deactivate never updates `provisionerStatus`, so a disconnected request can still show `provisioned` plus `endpoint`/`spec`.

If that's intentional, worth calling out in the API contract. Otherwise this should move to a terminal status (or clear the Information fields) on delete.

provisioner-api/src/clients/feed/types.ts
```diff
@@ -25,3 +25,29 @@ export interface Application {
   description?: string;
   namespace: string;
 }
+
+export interface ProvisionerStatus {
+  status: 'pending' | 'provisioned' | 'failed';
+  message?: string;
+  [information: string]: unknown;
```
Elson9, line 34: This index signature lets `endpoint` / `spec` persist, but the published SDX models still only allow `status` and `message` (`src/services/batch/types.ts` `ProvisionerStatus`, and `ConnectionRequestInput` in `src/controllers/sdx/v1/types.ts`). SDX v1 is generated with `noImplicitAdditionalProperties: throw-on-extras`, so clients using the OpenAPI spec can drop those fields.

If the goal is for clients to read this data, please extend the public `ProvisionerStatus` type and regenerate the spec.

Elson9 (changes requested):

src/server.ts
ikethecoder: Just to clarify, yes /feed skips the oauth proxy so within our namespace it is callable, but it isn't public as the API Gateway blocks traffic to /feed.

src/batch/feed-worker.ts
phowells: That makes sense. I’ve replaced the `ConnectionRequest`-specific handler with a generic `listFeedWorker` on `/feed/:entity`. It gets the query and fields from `data-rules`, then uses the configured transformations to decode JSON-backed values. This keeps the datafix on the internal feed path without adding entity-specific logic to the worker.

src/server.ts
phowells: Thanks for flagging this. Aidan’s clarification changes the picture here: `/feed` is blocked at the API Gateway and is only reachable inside the namespace, which is the existing trust boundary for this integration. I’m keeping that internal route, but the handler is now generic rather than being tied to `ConnectionRequest`.

provisioner-api/src/datafixes/aps-4790.ts
phowells: Good catch. I’ve changed the datafix so `pending` remains eligible on a later run; only `provisioned` and `failed` count as finished. The tests now cover object values as well as the JSON strings read from storage.

provisioner-api/src/controllers/connections-controller.ts
phowells: This can definitely happen on a transient Feed API failure. I’ve added three attempts with a short backoff for network errors, 408/429 responses, and 5xx responses, always retrying the same status value. If those attempts are exhausted, the error still bubbles up. Recovering after a process dies or the Feed API stays unavailable needs durable queue or reconciliation support, so I’d treat that as a separate architectural change rather than hide the failure here.

provisioner-api/src/controllers/connections-controller.test.ts
phowells: That behavior is intentional: `provisionerStatus` records the most recent provisioning attempt, while `isActive` is the source of truth for whether the connection is active now. I’ve added that distinction to the endpoint documentation, so deactivation keeps the last status and its information fields instead of introducing another state.

provisioner-api/src/clients/feed/types.ts
phowells: Agreed. `endpoint` and `spec` are now explicit in the public `ProvisionerStatus` models, the SDX connection input, the data-rule metadata, and the provisioner client types. I also regenerated the SDX v1 spec locally to confirm both fields appear while additional properties remain disabled.

ikethecoder (approved):

Elson9 (approved):

Issue #83, opened 2021-06-22
Provide contact information in the Reject email


ikethecoder:
When an Access Manager rejects a request, the email should include the Contact details (defined at the namespace) so that the requester could have a way of enquiring about the rejection.

ikethecoder:
Should it be the Email and Name of the Access Manager that rejected the request?

PR #1546, opened 2026-08-18
APS-4720: Add state-aware sdx-keys.r1 rotation with random kids

Elson9:
## Summary
- Extend `sdx-keys.r1` with optional `operation: add | rotate | replace | delete` while keeping legacy no-operation `{urn}:0` behavior.
- Fetch current keyset from GWA, emit complete desired state, and return structured non-secret `changes` for activity logging.
- Stop hardcoding edge `:0` kids in p2p/service patterns (`keyset_name` instead). Org `trust-kms` kids stay on `:0`.

## Test plan
- [x] Legacy no-`operation` still publishes `:0`
- [x] add / rotate / replace / targeted delete, plus last-key delete rejected
- [x] Retries with the same public key reuse the existing kid
- [x] Catalog activity records added/rotated/replaced/removed without key material (unit)
- [x] Catalog Cypress lifecycle (`08-catalog-activity` 19/19 local and CI; `gwa-api` #293 and `kong-oss-plugins` #83 are on `dev` / `feature/mtls`)

---
🚀 Feature branch deployment: https://api-services-portal-feature-aps-4720-sdx-keys.apps.silver.devops.gov.bc.ca


phowells (commented):
@Elson9, could you rebase this onto the latest `dev`? GitHub is showing merge conflicts at the moment.

Cypress is also red. Five failures are in the familiar `14-org-assignment` flow, but these two look related to this PR and were failing on both runs:

- `21-sdx-api/v1/08-catalog-activity.ts` — records added activity on initial runtime-group key apply
- `21-sdx-api/v1/09-e2e-tests.ts` — PUT `/organizations/{org}/connections`

Could you take a look at those when you rebase? I’d rather review the updated diff once the branch and relevant tests are clean.

Elson9:
> @Elson9, could you rebase this onto the latest `dev`? GitHub is showing merge conflicts at the moment.
> 
> Cypress is also red. Five failures are in the familiar `14-org-assignment` flow, but these two look related to this PR and were failing on both runs:
> 
> * `21-sdx-api/v1/08-catalog-activity.ts` — records added activity on initial runtime-group key apply
> * `21-sdx-api/v1/09-e2e-tests.ts` — PUT `/organizations/{org}/connections`
> 
> Could you take a look at those when you rebase? I’d rather review the updated diff once the branch and relevant tests are clean.

@phowells Rebased onto latest `dev`. Locally, `14-org-assignment` and `08-catalog-activity` are green. `09-e2e-tests` still 401s on the dataplane ping because Kong’s `feature/mtls` image doesn’t support `trust-sign` `keyset_name` yet.

CI needs these two APS-4720 PRs first:

- [kong-oss-plugins#83](https://github.com/bcgov/kong-oss-plugins/pull/83) → `feature/mtls` (unblocks 09)
- [gwa-api#293](https://github.com/bcgov/gwa-api/pull/293) → `dev` (unblocks 08 in CI)

ikethecoder:
Few initial observations:

- `operation=add` (or rotate) should probably be the default, as that is the desired outcome - currently if operation is excluded, it creates the old kid, which we want to move away from.
- I did a `operation=rotate` and it succeeded but it didn’t do any kong config update - was that because it detects that the certificate is the same?  Maybe an error if rotating the same key?
- I did a `operation=delete` and it gave a 500 response (looks like the provisioner error is not flowing back through the Portal cleanly
- I did a `action=delete` with an operation specified, and it gave a 500 - when the operation was not specified, it did the deletion ok.
- the `kid` looks a bit odd and long `urn:ca:bc:sdx:edge:rg0:dev:8875a149-2cd6-4f0d-aa8a-02ba793c7f5b` - wondering whether we should just use the uuid - or keep the urn but take a portion of the uuid (first 8 characters).  It just has to be unique in the context of the keyset.

ikethecoder:
Is there a functional difference between `add` and `rotate` as far as the gateway pattern is concerned?  Wondering if there is opportunity to simplify it a bit... maybe `add` is the default operation and `rotate` is removed as an option.  In the case of `add` - may make sense also to have a `max-keys` where the pattern will push out the older keys (LIFO) - not sure if key ordering is preserved (?).

Elson9:
@ikethecoder Thanks for the review. These are in now:

- Omitted `operation` on a runtime group defaults to `add`. Org/subsystem still publish `:0` (`trust-kms` pins that kid).
- `rotate` of a key that’s already in the set is **422**. `add` of the same material does nothing.
- Provisioner 4xx now comes through the portal (last-key delete **422**; `action=delete` + `operation` **400**). Query `action=delete` with no `operation` still wipes the qualifier.
- New kids are the URN plus the first 8 hex digits, e.g. `urn:ca:bc:sdx:edge:rg0:dev:8875a149`. A full URN still works for targeting existing keys.

On dropping `rotate` and adding `max-keys` LIFO: `add` vs `rotate` currently differ (`add` is no-op if the cert is already there; `rotate` requires new material) so we can overlap (publish new → promote → grace → delete the old kid). Would `max-keys` drop the oldest key in the same apply? I’d rather not remove the outgoing kid before promote/grace, and the data plane matches the mounted key, not keyset order. Happy to collapse to `add` + explicit `delete` if that’s what you want, just confirming the retire step.

ikethecoder:
> @ikethecoder Thanks for the review. These are in now:
> 
> * Omitted `operation` on a runtime group defaults to `add`. Org/subsystem still publish `:0` (`trust-kms` pins that kid).
> * `rotate` of a key that’s already in the set is **422**. `add` of the same material does nothing.
> * Provisioner 4xx now comes through the portal (last-key delete **422**; `action=delete` + `operation` **400**). Query `action=delete` with no `operation` still wipes the qualifier.
> * New kids are the URN plus the first 8 hex digits, e.g. `urn:ca:bc:sdx:edge:rg0:dev:8875a149`. A full URN still works for targeting existing keys.
> 
> On dropping `rotate` and adding `max-keys` LIFO: `add` vs `rotate` currently differ (`add` is no-op if the cert is already there; `rotate` requires new material) so we can overlap (publish new → promote → grace → delete the old kid). Would `max-keys` drop the oldest key in the same apply? I’d rather not remove the outgoing kid before promote/grace, and the data plane matches the mounted key, not keyset order. Happy to collapse to `add` + explicit `delete` if that’s what you want, just confirming the retire step.

Yeah `max-keys` would drop the oldest - so if it was set to 3, it would include the transitional new k1 and old k2 plus a really old k3.  But I think its more correct that at the end of the rotation, just valid keys remain.  So agree keeping as you have with an explicit delete is better.

provisioner-api/src/clients/oauth.ts
```diff
@@ -93,6 +93,7 @@ async function runClientCredentials(
       expiresIn: tokens.expires_in,
       scope: tokens.scope,
       accessToken: redactToken(tokens.access_token),
+      fullAccessToken: tokens.access_token,
```
phowells: Could we remove `fullAccessToken` before this merges? This writes the complete client-credentials bearer token to application logs on every issue or refresh. The redacted `accessToken` field immediately above already gives us useful troubleshooting context without placing a live GWA/SDX credential in log storage.

provisioner-api/src/services/gateway-patterns/sdx-keys.ts
```diff
@@ -235,21 +321,400 @@ export class SDXKeysPattern implements PatternProcessor {
...
+    const response = await this.gatewayAdmin.getKeys(
+      gatewayId,
+      environment,
+      tag,
+      keySetName
+    );
+
+    return (response.keys ?? [])
+      .map((key) => toExistingKey(key))
+      .filter((key): key is ExistingKey => key !== undefined);
```
phowells: I’m concerned about silently dropping entries here. This list becomes the complete desired keyset sent back to GWA, but an existing key with a missing name/kid, malformed JWK, unsupported key type, or unusable PEM is filtered out. On the next rotation that key is therefore absent from the deck—and from `changes.removed`—which could retire a still-active key unintentionally. Could we reject the operation if any returned key cannot be reconstructed, and cover that case in the state-aware tests?

docs/sdx-keys-rotation.md
```diff
@@ -0,0 +1,141 @@
...
+updated plugin image **before** applying those pattern changes.
+
+## Rotation procedure (runtime group)
+
+1. Create a one-time CA token:
+   `POST /organizations/{org}/runtime-groups/{name}/environments/{env}/tokens`
+2. Bootstrap / stage a **new** runtime-group key and CSR without restarting
+   Kong yet (`bootstrap.stageSecret=true`, which writes
+   `{release}-client-next` and skips `rollout restart`).
+3. Sign the CSR with the one-time token.
```
phowells: The chart’s staging Job already generates the key and CSR, then runs `step ca sign` with the one-time token; it stores only `tls.crt` and `tls.key`, so there isn’t a CSR left for a separate step 3. Could we fold signing into step 2 and have the next step tell the operator to extract `tls.crt` from the `*-client-next` Secret for `certificatePem`? That would keep this procedure aligned with the current `aps-devops#15` workflow.

phowells (changes requested):

provisioner-api/src/clients/oauth.ts
phowells: Confirmed—the raw fullAccessToken field is gone, while the redacted token summary remains for troubleshooting. That takes care of the credential-in-logs concern for me.

provisioner-api/src/services/gateway-patterns/sdx-keys.ts
phowells: This now stops the operation when any existing key cannot be reconstructed, and the mixed-set test covers the destructive case we were worried about. I’m satisfied this has been addressed.

docs/sdx-keys-rotation.md
phowells: The procedure now lines up with the chart: the staging Job signs the CSR, and the operator extracts tls.crt for certificatePem. Thanks—this closes the documentation gap.

phowells (approved):
Approved

ikethecoder (approved):

